PR #2, opened 2022-06-28
Docstrings RFC

sezna:
This RFC describes docstrings in the Sway language. See the RFC's content for more details 😄

### DIFF
--- a/rfcs/0000-docstrings.md
+++ b/rfcs/0000-docstrings.md
@@ -1,0 +1,88 @@
+- Feature Name: docstrings
+ - Start Date: 2020-06-27
+ - RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/2)
+ - Sway Issue: [FueLabs/sway#149](https://github.com/FuelLabs/Sway/issues/149)
+
+ # Summary
+ [summary]: #summary
+
+ Sway will convert Rust-style docstrings (`///`-style comments) into a `#[doc("documentation string")]` annotation. This annotation can then be consumed by tooling and plugins to produce
+ rendered documentation, a la `cargo doc` or `docs.rs`.
+
+ # Motivation
+ [motivation]: #motivation
+
+ As Sway's designs are largely motivated by Rust's design, we have always wanted to support in-code documentation in this manner. In-code documentation support has been part of the Sway vision
+ since the beginning.
+
+ This particular approach is both low-friction, in that it utilizes our existing features (annotations, comments, forc plugins); and consistent with Rust's implementation of docstrings. Upon
+ the implementation of this RFC in Sway, forc plugins should be able to produce documentation and documentation-related tooling based on in-code docstrings.
+
+ # Guide-level explanation
+ [guide-level-explanation]: #guide-level-explanation
+
+ A _docstring_ is a line of prose within a Sway source code file. This plain-language string serves to provide exposition and contextual explanation to the subsequent line of Sway code.
+
+ Docstrings, for convenience, may be written as a comment with three slashes. E.g.:
+
+ ```rust
+ /// The entry point to the script.
+ fn main() {}
+ ```
+
+ The above code example shows a _docstring_ documenting a function item, `fn main()`.
+
+ Generally, Sway programmers should find this to be the most expedient and convenient way to provide API documentation for libraries and contract ABIs, and documentation of internal concepts within scripts and predicates.
+
+ This style of docstring should be familiar to Rust programmers, and the concept of docstrings is generally prevalent in modern programming languages and should not require significant explanation or new educational material.
+
+ # Reference-level explanation
+ [reference-level-explanation]: #reference-level-explanation
+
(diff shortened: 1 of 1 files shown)


mohammadfawaz:
If we want to be strict with formatting, maybe we should enable a `markdown-link-check` CI check.

adlerjohn:
We should do that, especially if we're going to be getting external contributions. And not just linting. Here's the validator for EIPs: https://github.com/ethereum/eip_validator

rfcs/0000-docstrings.md
```diff
@@ -0,0 +1,91 @@
...
+have no canonical method of in-code documentation.
+
+
+# Prior art
+[prior-art]: #prior-art
+
+The obvious instance is Rust's docstrings. [Read more about that here](https://doc.rust-lang.org/rust-by-example/meta/doc.html).
+
+Additionally, many other languages have docstrings either via third party tooling or via native support:
+1. Javascript (`/**`-style)
```
adlerjohn, line 77: Newlines around (before and after) lists

rfcs/0000-docstrings.md
```diff
@@ -0,0 +1,101 @@
...
+
+Generally, Sway programmers should find this to be the most expedient and convenient way to provide API documentation for libraries and contract ABIs, and documentation of internal concepts within scripts and predicates.
+
+This style of docstring should be familiar to Rust programmers, and the concept of docstrings is generally prevalent in modern programming languages and should not require significant explanation or new educational material.
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Docstrings should be implemented as attributes (a.k.a. annotations) ([reference 1](https://github.com/FuelLabs/sway/issues/470), [reference 2](https://github.com/FuelLabs/sway/pull/1518), [reference 3 (rust)](https://doc.rust-lang.org/reference/attributes.html)). Because annotations/attributes can be applied to any `Item` in Sway ([reference 1](https://github.com/FuelLabs/sway/blob/master/sway-parse/src/attribute.rs#L4), [reference 2](https://github.com/FuelLabs/sway/blob/ba30e8e5ccbb0512aacbaee594473da9e0839c3d/sway-parse/src/item/mod.rs#L13)), this means that any `Item` can be documented with this feature.
```
mitchmindtree, line 46: This would also be a great reference to keep track of:

https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html

mitchmindtree (approved):
Lgtm!

Fwiw, I recently added some rough thoughts on how we could go about implementation [here](https://github.com/FuelLabs/sway/issues/161#issuecomment-1153659295). This RFC appears to cover steps 1 and 2 specifically, which sounds like a great start.

Very happy to go with the `#[doc("docstring")]` approach over Rust's `#[doc = "docstring"]`, especially seeing as we already have `#[attr(args)]` support anyway and there doesn't appear to be any great motivation for adding assignment when we can just nest attributes.

---

## Future `#[doc]` features

While probably out of scope of this initial/minimal RFC, we should also keep in mind that Rust's `#[doc]` attributes also allow for doing other fancy things like specifying favicons/logos, inlining documentation from one module into another (particularly useful when re-exporting docs from another module e.g.), aliases and more.

That said, I think all of these features could be proposed later and safely added on top. E.g. The following:

```rust
/// Blah blah blah
#[doc(alias("Bar"))]
struct Foo {}
```
would just get desugared to something like:
```rust
#[doc(alias("Bar"), "Blah blah blah")]
struct Foo {}
```
Where the string literal at the `doc`-level is always the doc string, and each of the other features are specified by a nested attribute.

otrho (approved):

rfcs/0000-docstrings.md
```diff
@@ -0,0 +1,101 @@
+- Feature Name: docstrings
+- Start Date: 2020-06-27
```
mohammadfawaz, line 2:
```suggestion
- Start Date: 2022-06-27
```

rfcs/0000-docstrings.md
```diff
@@ -0,0 +1,101 @@
...
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/2)
+- Sway Issue: [FueLabs/sway#149](https://github.com/FuelLabs/Sway/issues/149)
+
+# Summary
+
+[summary]: #summary
+
+Sway will convert Rust-style docstrings (`///`-style comments) into a `#[doc("documentation string")]` annotation. This annotation can then be consumed by tooling and plugins to produce
+rendered documentation, a la `cargo doc` or `docs.rs`.
```
mohammadfawaz, line 11: These end lines make the text a bit harder to read and review. I personally like the `gq` command in Vim for consistency (even though it adds more end lines), but a single long line is also fine. Totally up to you though :)

rfcs/0000-docstrings.md
```diff
@@ -0,0 +1,101 @@
...
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+In the space of possible docstring designs, this is the most consistent with Rust and also very ergonomic. Additionally, a good chunk of Sway code has already been written with the assumption that this docstring format will be accepted. Therefore, there are already many docstrings written in this way.
+
+Other designs could include just the attribute without the three-slash-comment-based syntactic sugar, ML-family-style docstrings `{- docstring -}`, or multiline-style:
+
+```
+**
```
mohammadfawaz, line 66: Is this the correct way?
```suggestion
/**
```

mohammadfawaz (changes requested):

mohammadfawaz (approved):

tritao (approved):

otrho (approved):

mitchmindtree (approved):

emilyaherbert (approved):

PR #8, opened 2022-07-04
Change the behaviour of paths and namespace resolution

canndrew:
Overhaul the way paths and namespaces work in Sway, as per (https://github.com/FuelLabs/sway/issues/2125).

[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/canndrew/paths-rfc/rfcs/0000-paths.md)

### DIFF
--- a/rfcs/0000-paths.md
+++ b/rfcs/0000-paths.md
@@ -1,0 +1,302 @@
+- Feature Name: specify_path_resolution
+- Start Date: (fill me in with today's date, YYYY-MM-DD)
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/001)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes a change to the way paths and modules work in Sway.
+
+Paths are sequences of `::`-delimited idents which are used to refer to a
+specific item in a specific module. Paths can either be absolute or relative
+depending on whether they start with a `::` token.
+
+Under this proposal the first identifier in an absolute path is always the name
+of a package. This package either could be a dependency declared in the
+package's Forc.toml file, or the name of the current package. In contrast,
+relative paths are always relative to the current module.
+
+As part of the transition to this new system, this RFC also proposes to replace
+`dep` declarations with `mod` declarations and to deprecate the use of program
+kind headers at the start of Sway source files.
+
+# Motivation
+
+[motivation]: #motivation
+
+There are currently several issues with the way modules and packages are named
+and referred to in Sway.
+
+* The name a library module gives itself does not need to match the name given
+  in the `dep` declaration which includes it.
+* The name the root library module of a library package gives itself does not
+  need to match the name of the package.
+* Due to the above two issues, its possible to have namespace collisions where
+  two separate packages or modules can be imported despite having the same
+  name.
+* It's possible for modules under the root module of package to have name
+  conflicts with dependency packages.
+
+With the new rules proposed in this RFC, all modules that are in scope in (both
+those in the current package and those in dependency packages) have a distinct
+and unique absolute path. What's more, relative paths are always unambiguous
+and refer to a specific item in a specific module.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Program kinds in Forc.toml
+
+The `[project]` section of a package's Forc.toml is required to have a `kind`
+field specifying the kind of the program. This value of this field must be one
+of `"library"`, `"contract"`, `"script"` or `"predicate"`. This replaces the
+now-deprecated program kind header at the start of a Sway source file.
+
+Since a package's Forc.toml also specifies the name of the package this change
+de-duplicates the library name which was previously also given by the `library
+name;` program kind header.
+
+## `mod` declarations
+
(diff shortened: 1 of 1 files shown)


mitchmindtree (commented):
After having worked with `Namespace` quite a bit recently, I'm a big fan of this proposal :)

I've left some thoughts, mostly just agreeing / mentioning some future considerations.

The only actionable change I'd like to see in this RFC is to pick one approach for parent module file hierarchy support, e.g. either:

1. `foo/mod.sw` or
2. `foo.sw` with matching `foo/` directory

but not both.

At such an early stage we have the opportunity to pick "the way".

canndrew:
> In Rust, I personally prefer the foo/mod.rs approach to keep module branches isolated under 1 directory. However, for Sway I'd much prefer we have a single, consistent approach.

Ditto actually. I'd rather we just have one method, and I'd rather that method was `foo/mod.sw`. I guess I should have made the RFC more opinionated.

> I think if we want std and core to be available via their names by default (e.g. so that you can do std::hash::sha256 without having to use ::std first), we can handle that as a part of providing a prelude when the time comes.

Agreed. As another future consideration, something that comes up in Rust-land occasionally is the idea of having crate-wide custom preludes. In all my Rust programs I've taken to defining a `priv_prelude` module which imports all the stuff that I want to be able to use throughout the library and then just putting `use crate::priv_prelude::*` at the start of each module. This avoids having to have the same pile of `use` statements in each module, makes sure that the names are available everywhere and that they refer to the same thing everywhere. So it might be nice to have first-class support for this pattern (though I don't have an idea of what that would look like).

Relatedly, something I really dislike in Rust code is having multiple glob-imports in a single module. I use LSP, but I still like being able to see where things are imported from at the top of the file. If there's a single glob import then I know that anything not explicitly imported came from there, but if there's multiple then I'm forced to use the language server to find where things are coming from. So I think it'd be nice if glob imports were restricted to only allow one per module (though I can imagine scenarios where this might be too restrictive).

otrho:
> I'd rather we just have one method, and I'd rather that method was `foo/mod.sw`

Oooh, I prefer the new `foo.sw`. 🙂 Maybe we'll have to vote on it.

canndrew:
> Oooh, I prefer the new `foo.sw`.

The most annoying thing about the `foo.sw` convention is that it breaks tab-completion, both in vim and from the shell.

Suppose I'm navigating to where I expect to find `foo::bar`

With `foo/mod.sw`:
    * Start typing `foo`.
    * Hit tab.
    * If it doesn't complete to `foo.sw` or `foo/` then type more characters and tab again.
    * If it still doesn't complete then I'm being dumb and I made a typo or I'm in the wrong directory or something.
    * If it completes to `foo.sw` then hit enter.
    * Otherwise if it completes to `foo/` hit tab again.
    * See if any files match what I'm looking for.
    * Start typing the submodule name or `mod.sw`.
    * Hit tab and enter again.

With `foo.sw`:
    * Start typing `foo`.
    * Hit tab.
    * It completes to `foo`.
    * I don't know whether I need to type more characters because something else starts with `foo` or not.
    * Tentatively hit `/` and tab again.
    * If there's no subdirectory by that name then it doesn't work and I have to hit backspace.
    * Otherwise if there's a subdirectory but none of the files look like what I'm after hit backspace again.
    * Type `.sw` and enter.
    * If I'm being dumb and I made a typo or I'm in the wrong directory or something then I just opened a non-existent file.

This has been my experience working on the Sway codebase where we use the `foo.sw` convention at least. It's a small annoyance but it accumulates when you have to do it a hundred times a day. Maybe I just need to reprogram my muscle memory though.

Also having a file and directory with the same name rubs me the wrong way, duplicated state and all that.

> Maybe we'll have to vote on it.

That's an option. I'd prefer either way over both :)

otrho:
Hrmm.  If you want to open the module `foo` then it will be in `foo.rs` and its submodules will be under `foo/`.  So `f`, tab to `foo`, hit `.` if I want `foo.rs`  or `/` if I don't.   It's only really a problem if you don't know what you're looking for.

The argument against `foo/mod.rs` is the usual one of looking at the name of the current buffer and seeing `mod.rs` and not knowing which module you're actually looking at, or wanting to switch to the `foo` module but the buffer is called `mod.rs` just like 3 others which are also open.  I also like it that if a module doesn't have submodules then you don't need a subdirectory at all, rather than a subdirectory with one file in it.

rfcs/0000-paths.md
```diff
@@ -0,0 +1,302 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Program kinds in Forc.toml
+
+The `[project]` section of a package's Forc.toml is required to have a `kind`
+field specifying the kind of the program. This value of this field must be one
+of `"library"`, `"contract"`, `"script"` or `"predicate"`. This replaces the
+now-deprecated program kind header at the start of a Sway source file.
```
adlerjohn, line 56: Thought: currently, a project can mix in libraries with a contract/script/predicate. After this change, they cannot. Not inherently a bad thing, especially after workspaces are introduced. Just something to note.

emilyaherbert:
I know attention is elsewhere right now, but I am just commenting this now so that I don't forget---when there is movement on this again, would you be able to provide a sample Sway repo with code examples of what everything might look like with your proposed changes @canndrew ? The code wouldn't compile obviously, but I think it would give me a better feel for how the proposed changes would play out.

sezna:
Closing as this is not on the roadmap and is not going to be actively worked on. Feel free to reopen in the future if we deem this necessary.

PR #9, opened 2022-07-07
RFC: Calling context attribute 

tritao:
Adds support for an attribute (i.e. `#[context(internal_only)]`) to annotate the supported calling contexts for particular functions.

This annotation is then be used by the compiler to check and restrict the calling of these functions to a compatible calling context.

[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/calling-context-attribute/rfcs/0000-rfc-calling-context-attribute.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,8 @@
 # Sway RFCs
 
-
-
 ## Active RFCs
-_None_
+* [Docstrings]
+
+| RFC                                                                                    |     Title          |
+| -------------------------------------------------------------------------------------- | ------------------ |
+| [0002](https://github.com/FuelLabs/sway-rfcs/blob/master/rfcs/0002-docstrings.md)      | Docstring Comments |
--- a/rfcs/0002-docstrings.md
+++ b/rfcs/0002-docstrings.md
@@ -1,0 +1,101 @@
+- Feature Name: docstrings
+- Start Date: 2022-06-27
+- RFC PR: [FuelLabs/sway-rfcs#0002](https://github.com/FuelLabs/sway-rfcs/pull/2)
+- Sway Issue: [FueLabs/sway#149](https://github.com/FuelLabs/Sway/issues/149)
+
+# Summary
+
+[summary]: #summary
+
+Sway will convert Rust-style docstrings (`///`-style comments) into a `#[doc("documentation string")]` annotation. This annotation can then be consumed by tooling and plugins to produce
+rendered documentation, a la `cargo doc` or `docs.rs`.
+
+# Motivation
+
+[motivation]: #motivation
+
+As Sway's designs are largely motivated by Rust's design, we have always wanted to support in-code documentation in this manner. In-code documentation support has been part of the Sway vision
+since the beginning.
+
+This particular approach is both low-friction, in that it utilizes our existing features (annotations, comments, forc plugins); and consistent with Rust's implementation of docstrings. Upon
+the implementation of this RFC in Sway, forc plugins should be able to produce documentation and documentation-related tooling based on in-code docstrings.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+A _docstring_ is a line of prose within a Sway source code file. This plain-language string serves to provide exposition and contextual explanation to the subsequent line of Sway code.
+
+Docstrings, for convenience, may be written as a comment with three slashes. E.g.:
+
+```rust
+/// The entry point to the script.
+fn main() {}
+```
+
+The above code example shows a _docstring_ documenting a function item, `fn main()`.
+
+Generally, Sway programmers should find this to be the most expedient and convenient way to provide API documentation for libraries and contract ABIs, and documentation of internal concepts within scripts and predicates.
+
+This style of docstring should be familiar to Rust programmers, and the concept of docstrings is generally prevalent in modern programming languages and should not require significant explanation or new educational material.
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+Docstrings should be implemented as attributes (a.k.a. annotations) ([reference 1](https://github.com/FuelLabs/sway/issues/470), [reference 2](https://github.com/FuelLabs/sway/pull/1518), [reference 3 (rust)](https://doc.rust-lang.org/reference/attributes.html)). Because annotations/attributes can be applied to any `Item` in Sway ([reference 1](https://github.com/FuelLabs/sway/blob/master/sway-parse/src/attribute.rs#L4), [reference 2](https://github.com/FuelLabs/sway/blob/ba30e8e5ccbb0512aacbaee594473da9e0839c3d/sway-parse/src/item/mod.rs#L13)), this means that any `Item` can be documented with this feature.
+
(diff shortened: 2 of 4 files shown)


otrho:
Sounds good to me.  I personally like `calling_context(contract_only)`.  `internal` in particular is a bit ambiguous.

tritao:
Updated the RFC with `calling_context(contract_only)` syntax instead which I also think its more explicit.

otrho (approved):

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+
+A _calling context attribute_ is an attribute that can be used to annotate function items in Sway to restrict the context in which they can be called.
+
+Take the following example:
+
+```rust
+contract;
+
+#[calling_context(contract_only)]
+fn bar() {}
```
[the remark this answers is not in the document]
tritao: I like your suggestions, definitely clears up and makes it explicit the behaviour for each program kind, my next revision of the RFC will have this in.

tritao:
So heads up, we had some discussion about this on yesterday's sync up call, specifically two points, around the transitivity of the attribute, where we concluded it does make sense for documentation purposes, and about extending the annotation to support external-only contexts as well, for an upcoming function that will be used to get the currently running script.

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+
+This annotation can then be used by the compiler to check and restrict the calling of these functions to a compatible calling context.
+
+# Motivation
+
+[motivation]: #motivation
+
+The main goal is to be able to get compile-time safety when working with functions that should only be called from a contract context.
+
+A real-world example of this is the [msg_amount() stdlib function](https://github.com/FuelLabs/sway-lib-std/blob/6a8f5bf588df5d0679fb834f05c900f2e54de426/src/context.sw#L33-L38) which only returns a proper value when used in a contract context. In an external context is returns a well-defined but not semantically representative value, specifically 0, as explained by @adlerjohn in [FueLabs/sway#963](https://github.com/FuelLabs/sway/issues/963).
```
mitchmindtree, line 20: I'm curious to see some more real-world examples of cases where we'd like to use this attribute.

If cases like using `msg_amount` are the only kinds of cases we have in mind, perhaps as an alternative we should consider having users take `msg_amount` via a function argument instead?

E.g. rather than

```rust
#[calling_context(contract_only)]
fn bar() {
    let msg_amt = msg_amount();
    // ...
}
```

Library writers can do:

```rust
fn bar(msg_amt: u64) {
    // ...
}
```

Of course, in contract code, the caller would then need to call `bar` with the explicit argument `bar(msg_amount())` rather than just `bar()`, however the former is arguably better for the reader as they can see where the value is coming from, rather than having `bar` introduce the side-effect?

I'm weary that our current motivations might not fully justify introducing a whole new attribute, but perhaps a more motivating example might help?

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+
+The attributes are parsed, and during type checking, we keep track of the current calling context for a particular type check context. Then during function and method applications, we can check if a particular function is missing a context attribute, and issue an error if that is the case.
+
+Later, as we do for purity, once the entire program is type-checked, we perform an analysis to figure out which top-level declarations are being called from the wrong context and issue the corresponding compiler error.
+
+The feature is pretty self-contained and doesn't affect any pre-existing code.
+
+The only expected potential breaking changes are once the standard library code will take advantage of this, then user code will need to be annotated further up the call chain.
+
+One thing to note is that is that functions that call an `contract_only` function needs themselves to be annotated as contract-only.
```
sezna, line 69:
```suggestion
One thing to note is that is that functions that call a `contract_only` function needs themselves to be annotated as contract-only.
```

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+fn foo() {}
+
+fn bar() {
+  foo();
+}
+```
+
+This will give a compile error, forcing the user to annotate `bar` with `#[calling_context(contract_only)]` as well.
+
+The only exception to this are contract-level functions which are themselves already satisfy the contract-only contraints by default.
```
sezna, line 84:
```suggestion
The only exceptions to this are contract-level functions which are themselves already satisfy the contract-only constraints by default.
```

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+We could go with an `contract_only` keyword I guess, but re-using attributes seems like a better approach since we are already using the same scheme for storage access.
+
+# Prior art
+
+[prior-art]: #prior-art
+
```
sezna, line 103: we also have some prior art [here](https://github.com/FuelLabs/sway/blob/master/sway-core/src/asm_generation/checks.rs#L32) where we manually check that certain opcodes are not being generated in the wrong context.

sezna: There's a benefit to catching that earlier as we have more ast information before these opcodes are generated, so we can improve error messages.

sezna (approved):

rfcs/0000-rfc-calling-context-attribute.md
```diff
@@ -0,0 +1,116 @@
...
+The obvious instance is Sway's very own storage attributes, which work used in a very similar manner, and from which the prototype implementation approach of this feature in the compiler was derived.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+1. Is `contract_only` the only calling context we want to support or can you think of some other we may want to support in the future?
```
adlerjohn, line 110: No, we also need predicate-only and script+contract-only (i.e. not-predicate).

- Predicate-only would be something like "get currently verifying predicate."
- Contract-only would be "read/write storage slot."
- Script+contract-only would be something like "contract call."

rfcs/0000-rfc-calling-context-attribute.md
sezna: Whoops, looks like this conversation was never resolved.

If we always defer the acquisition of the side-effectual information via arguments, we still eventually have to write functions in the standard library which produce this information and we want to gate access to them at compile time. This _could_ be accomplished with conditional compilation, but the error messages for that would not be great. We could improve them with conditional compilation _and_ a `compile_error!()` macro like Rust has?

E.g.

```rust
// in stdlib
#[cfg(not(external))]
fn msg_amount() {
  compile_error!("Message amount cannot be called from a non-external context.");
}

```

mitchmindtree: I can see us eventually wanting conditional compilation for sure, however this issue in particular (and potentially this entire RFC) might be better solved by https://github.com/FuelLabs/sway/issues/2463.

That is, if the contract-specific functions were moved to contract methods, it would be impossible to call them in an incorrect context. This way we have the added benefit that our contract-only functions also are not side-effecting, and it's always clear when a function might behave differently as a result of referencing some contract or msg state, as either the contract or the data itself would need to be passed as an argument.

sezna:
I think this is going to be deferred for a bit as nothing is blocked by it. But we should definitely reconcile how we want to handle this w.r.t #2463

PR #11, opened 2022-07-13
Declaration engine

emilyaherbert:
[rendered](https://github.com/FuelLabs/sway-rfcs/blob/emilyaherbert/declartion-engine/rfcs/0003-declaration-engine.md)

[prototype](https://github.com/emilyaherbert/declaration-engine-and-collection-context-demo/tree/master/de)


rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
+- Feature Name: Declaration Engine
+- Start Date: 2022-07-13
+- RFC PR: [FuelLabs/sway-rfcs#0011](https://github.com/FuelLabs/sway-rfcs/pull/11)
+- Sway Issue: [FueLabs/sway#1821](https://github.com/FuelLabs/sway/issues/1821), [FueLabs/sway#1692](https://github.com/FuelLabs/sway/issues/1692)
+
+# Summary
+
+[summary]: #summary
+
+With it's current design, the Sway compiler faces challenges regarding how declarations interact with the type system and code generation. These include:
```
sezna, line 10:
```suggestion
With its current design, the Sway compiler faces challenges regarding how declarations interact with the type system and code generation. These include:
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+With the changes introduced by this RFC, the compiler will be able to think about declarations more abstractly and will not be required to inline AST nodes during type checking. In addition to solving the issues above, I believe that this change will create a mechanism in the compiler which will allow us to implement additional optimizations in the future.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Because this is an internal compiler RFC, the affected group is the Sway compiler devs.
+
+Specifically in the context of Sway terminology, the phrase "type checking" is often used to refer to the Sway compiler step in which untyped AST nodes are transformed to typed AST nodes. But really, the compier step we refer to "type checking" is actually doing type inference. This distinction is important because it allows us to understand the role of the declaration engine.
```
sezna, line 29:
```suggestion
Specifically in the context of Sway terminology, the phrase "type checking" is often used to refer to the Sway compiler step in which untyped AST nodes are transformed to typed AST nodes. But really, the compiler step we refer to "type checking" is actually doing type inference. This distinction is important because it allows us to understand the role of the declaration engine.
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Toy Example
+
+(this is an elaborate explanation that most will already be profficient in---but it does build to a good point!)
```
sezna, line 33:
```suggestion
(this is an elaborate explanation that most will already be proficient in---but it does build to a good point!)
```

rfcs/0003-declaration-engine.md
sezna, line 29: I believe what we refer to as semantic analysis is performing _both_ type checking and type inference, right? Not just type inference.

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+For the purposes of illustration, imagine a new toy language that is statically and strongly typed, but allows for optional type annotations (like Sway and Rust). The toy compiler that we could write for this toy language would need to evaluate the types of objects at compile time to either correctly compile a program or fail to compile a program (like the Sway compiler and the Rust compiler). The toy compiler would have three broad-strokes steps (1) parsing, (2) type checking, and (3) code generation. Because the toy language allows for optional type annotations, after the parsing step, the toy compiler would know the types for either none, some, or all, of the objects in the program. To compensate for this, the toy compiler would perform type inference, where the types of untyped objects would be inferred from the types of typed objects.
```
sezna, line 35: I think `object` here is an ill-defined term, perhaps we  use the word item or expression? Or define what an object is?

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Because this is an internal compiler RFC, the affected group is the Sway compiler devs.
+
+Specifically in the context of Sway terminology, the phrase "type checking" is often used to refer to the Sway compiler step in which untyped AST nodes are transformed to typed AST nodes. But really, the compier step we refer to "type checking" is actually doing type inference. This distinction is important because it allows us to understand the role of the declaration engine.
+
+## Toy Example
```
sezna, line 31: Since this isn't really an example of what you are proposing, and is rather a description of a related concept, maybe we could title this section differently.
```suggestion
## Background Information: Sway's Type Inference
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+(Type inference, constraint solving, and unification are defined here: https://papl.cs.brown.edu/2014/Type_Inference.html)
+
+## Declaration Engine
+
+The declaration engine takes the concept of constraints and unification and applies it to declarations.
+
+For example, when type checking a function application, the compiler does not need to care about the contents of the body of the function, instead it is sufficient to create a constraint referring to "the return type of the function".
+
+For example, given a typed function declaration `add`, we add this declaration to the declaration engine. Then when type checking a function application of `add`, instead of inlining the function body, we create constraints referring to "the body of `add`" and "the return type of `add`". When type checking a _function application_
```
sezna, line 72:
```suggestion
For example, given a typed function declaration `add`, we add this declaration to the declaration engine. Then when type checking a function application of `add`, instead of inlining the function body, we create constraints referring to "the body of `add`" and "the return type of `add`". When type checking a _function application_.
```

sezna: Would the type of the body of the function not be the same as the return type of the function? I would think that the only constraints that would matter would be those of the function inputs, the type parameters, and then that of the function return type.

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+### Definition
+
+During type checking of a declaration, the compiler will add information about that declaration to the declaration engine:
+
+```rust
+struct DeclarationEngine {
+    // function name to typed function declaration
+    // this example code does not consider Path's, but we could use
+    // a BTreeMap with Path's as the key
+    functions: HashMap<String, TypedFunctionDeclaration>,
```
sezna, line 89: my 2 cents: I prefer strong types like `FunctionName`, so we can modularly define the key type later

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+    // struct name to typed struct declaration
+    // again, this does not consider Path's
+    structs: HashMap<String, TypedStructDeclaration>,
```
sezna, line 93: ditto re strongly typed key, `StructName` or similar

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+```rust
+struct DeclarationEngine {
+    // function name to typed function declaration
+    // this example code does not consider Path's, but we could use
+    // a BTreeMap with Path's as the key
+    functions: HashMap<String, TypedFunctionDeclaration>,
+
+    // struct name to typed struct declaration
+    // again, this does not consider Path's
```
sezna, line 92:
```suggestion
    // again, this does not consider Paths
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+### Definition
+
+During type checking of a declaration, the compiler will add information about that declaration to the declaration engine:
+
+```rust
+struct DeclarationEngine {
+    // function name to typed function declaration
+    // this example code does not consider Path's, but we could use
+    // a BTreeMap with Path's as the key
```
sezna, line 88:
```suggestion
    // a BTreeMap with Path's as the key
```
```suggestion
    // this example code does not consider Paths, but we could use
    // a BTreeMap with Paths as the key
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+    functions: HashMap<String, TypedFunctionDeclaration>,
+
+    // struct name to typed struct declaration
+    // again, this does not consider Path's
+    structs: HashMap<String, TypedStructDeclaration>,
+
+    // enums and traits
```
sezna, line 95: TODO?

sezna: maybe say `omitted` or something?

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+    fn add_struct(&mut self, r#struct: TypedStructDeclaration) -> DeclarationDefinition {
+        // add to self and return DeclarationDefinition
+    },
+
+    // enums and traits
+}
+
+enum DeclarationDefinition {
+    // does not consider Path's
```
sezna, line 111:
```suggestion
    // does not consider Paths
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+    Function(String),
+
+    // does not consider Path's
```
sezna, line 114:
```suggestion
    // does not consider Paths
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+    Struct(String),
+
+    // enums and traits
+}
+```
+
+Then in the typed AST, that declarations could look like (in pseudocode):
```
sezna, line 121:
```suggestion
Then in the typed AST, those declarations could look like (in pseudocode):
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+```rust
+contract;
+
+/// imports
+
+DeclarationDefinition::Function(<function name>)
```
sezna, line 128: perhaps this code is too pseudo for me, I don't quite understand it. Are these are representing imports?

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+DeclarationDefinition::Function("main")
+
+```
+
+Because inlining is still needed for the IR, these instances of `DeclarationDefinition` would be replaced in an additional step at the end of type checking called the [resolution step](https://github.com/FuelLabs/sway/issues/1820).
+
+### Usage
+
+When a declaration is _used_ an internal reference is created that references the existence of a possible declaration in the declaration engine, and that is resolved during the resolution step. Given:
```
sezna, line 141:
```suggestion
When a declaration is referred to, a reference to the existence of a possible declaration in the declaration engine is generated. That is resolved during the resolution step. Given:
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+}
+```
+
+The `TypedFunctionDeclaration` for `add` is added to the declaration engine. Then given:
+
+```rust
+let foo = add(1, 2);
+```
+
+`foo` will be a `TypedExpression` with a [`TypedExpressionVariant::FunctionApplication`](https://github.com/FuelLabs/sway/blob/dcaee960df9b3466c6897d0a86ea806c9abf7edd/sway-core/src/semantic_analysis/ast_node/expression/typed_expression_variant.rs#L18), where instead of storing the entire `TypedFunctionDeclaration` in the `TypedExpressionVariant::FunctionApplication` variant, there is a reference created `Function("add", vec!(u64, u64))` which is stored [here](https://github.com/FuelLabs/sway/blob/dcaee960df9b3466c6897d0a86ea806c9abf7edd/sway-core/src/semantic_analysis/ast_node/expression/typed_expression_variant.rs#L23). The type of `foo` is some new `TypeInfo::ReturnType("add", vec!(u64, u64))`. Both the reference `Function("add", vec!(u64, u64))` and `TypeInfo::ReturnType("add", vec!(u64, u64))` are resolved during the resolution step.
```
sezna, line 155:
```suggestion
`foo` will be a `TypedExpression` with a [`TypedExpressionVariant::FunctionApplication`](https://github.com/FuelLabs/sway/blob/dcaee960df9b3466c6897d0a86ea806c9abf7edd/sway-core/src/semantic_analysis/ast_node/expression/typed_expression_variant.rs#L18), where instead of storing the entire `TypedFunctionDeclaration` in the `TypedExpressionVariant::FunctionApplication` variant, there is a reference created `Function("add", vec![u64, u64])` which is stored [here](https://github.com/FuelLabs/sway/blob/dcaee960df9b3466c6897d0a86ea806c9abf7edd/sway-core/src/semantic_analysis/ast_node/expression/typed_expression_variant.rs#L23). The type of `foo` is some new `TypeInfo::ReturnType("add", vec![u64, u64])`. Both the reference `Function("add", vec![u64, u64])` and `TypeInfo::ReturnType("add", vec![u64, u64])` are resolved during the resolution step.
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+Importantly, introducing the declartion engine in this way means that during type checking before the resolution step, the references introduced during usages of declarations are assumptions. There is no check happening to ensure that `add` exists when creating `Function("add", vec!(u64, u64))`. The check to see if something exists happens in the resolution step. Once the _collection context_ is introduced, the concept of checking to see if things exist can happen significantly sooner, even before type checking begins at all, if we wanted to do so.
```
sezna, line 157:
```suggestion
Importantly, introducing the declaration engine in this way means that during type checking before the resolution step, the references introduced during usages of declarations are assumptions. There is no check happening to ensure that `add` exists when creating `Function("add", vec!(u64, u64))`. The check to see if something exists happens in the resolution step. Once the _collection context_ is introduced, the concept of checking to see if things exist can happen significantly sooner, even before type checking begins at all, if we wanted to do so.
```

sezna, line 157:
```suggestion
Importantly, introducing the declartion engine in this way means that during type checking before the resolution step, the references introduced during usages of declarations are assumptions. There is no check happening to ensure that `add` exists when creating `Function("add", vec!(u64, u64))`. The check to see if something exists happens in the resolution step. Once the _collection context_ is introduced, the concept of checking to see if things exist can happen earlier in compilation, even before type checking begins at all, if we wanted to do so.
```

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+I feel that this design offers the best compromise between current state and future growth. The concept of a "declaration engine" in general is very broad, which is a good thing! The design in this RFC is a good one to get us started, but I imagine that as the language progresses, it's needs and use cases for the declaration engine will expand and change. Introducing the declaration engine now gives us the opportunity to create a mechanism through which some of those changes could be introduced later down the road.
+
+That being said, from a high-level perspective, the alternative design would be something similar to what we are doing now, which is more focused on inlining things.
+
+# Prior art
+
+[prior-art]: #prior-art
+
+Well, I'm actually not sure. Off the top of my head, at a high-level there are languages that do not inline functions and instead use function tables, for instance. In those languages, I'm really not sure if this is done via a "declaration engine" or not. This RFC will not allow us to avoid inlining functions (we are still doing it in the IR), but it is a first step.
```
sezna, line 177: I believe we should do at least some research before diving in, given this is a multi month journey 😄  Here are some links I think may be good starting points:
1. [rustc item collection for monomorphization](https://github.com/rust-lang/rust/blob/master/compiler/rustc_monomorphize/src/collector.rs)
2. [rustc collection item context](https://github.com/rust-lang/rust/blob/master/compiler/rustc_typeck/src/collect.rs#L105)
3. maybe helpful, maybe not, as Idris is higher level...but [idris has a context it uses to resolve types](https://github.com/idris-lang/Idris2/blob/main/src/Core/Context/Context.idr#L417)
4. Maybe @tritao has thoughts here? It seems he wrote [something similar](https://github.com/tritao/CppSharp/blob/master/src/AST/ASTContext.cs) in his older project.

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+
+    // struct name to typed struct declaration
+    // again, this does not consider Path's
+    structs: HashMap<String, TypedStructDeclaration>,
+
+    // enums and traits
+}
+
+impl DeclarationEngine {
+    fn add_function(&mut self, function: TypedFunctionDeclaration) -> DeclarationDefinition {
```
sezna, line 99: If the function has to be typed when we insert it into the declaration, doesn't that mean we maintain the status quo of having to type check (and perform inference on) everything before beginning this process?

sezna (changes requested):

sezna:
Sooo excited for work on this to begin 🚀

rfcs/0003-declaration-engine.md
```diff
@@ -0,0 +1,189 @@
...
+DeclarationDefinition::Function(<function name>)
+DeclarationDefinition::Struct(<struct name>)
+
+/// storage declaration
+
+DeclarationDefinition::Function("main")
+
+```
+
+Because inlining is still needed for the IR, these instances of `DeclarationDefinition` would be replaced in an additional step at the end of type checking called the [resolution step](https://github.com/FuelLabs/sway/issues/1820).
```
otrho, line 137: Inlining isn't really needed for the backend and the resolution step you link to there is more about confirming that the AST is fully resolved/unified, no unknown types remain.

Is this a third definition of 'inlining'?  :slightly_smiling_face: 'Ensuring all generic types are replaced with concrete types'?

The back end just wants low level consistent fully-qualified naming for declarations and references.  So, for functions in a module it's easy -- `crate::module::fn_name` -- but for e.g., trait methods it's harder.  When we call `foo_bar.into()` we need a `crate::module::concrete_impl::into` name to call.

otrho (commented):
I don't think you've addressed out of order declarations and dependencies here?  Is that another thing the decl engine was supposed to handle?

Allowing out of order declarations and to remove the declaration sorting we have now _and_ especially to allow recursive data structures and function calls is highly desirable.

mitchmindtree (commented):
Nice one Emily!

I don't have much to add on top of the others' comments, though I think your insight about our current lack of clarity on how to perform _type inference_ in a consistent manner being a core issue is a really key insight and makes a lot of sense! I can see the proposed `DeclarationEngine` going a long way to help with this.

The described approach reminds me a little of [the Hindley-Milner type system](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system) and its approach to inference, though maybe a little less formal and jargon-y :smiling_face_with_tear: Considering it was the solution adopted by ML, Haskell, Rust and I'm sure many other languages, perhaps it's worth taking a closer look and seeing if there's anything we can learn from it as "Prior Art"? I'd imagine it might have some key insights particularly when it comes to handling trait (typeclass) constraints and polymorphic types (`Vec<T>`, `(T, U)`, etc).

sezna:
> The described approach reminds me a little of [the Hindley-Milner type system](https://en.wikipedia.org/wiki/Hindley%E2%80%93Milner_type_system) and its approach to inference,

I actually based our current type engine on Hindley-Milner when I wrote it, hence the function names like `unify` etc. We are already using the HM algorithms to perform unification and inference.

rfcs/0003-declaration-engine.md
emilyaherbert: Good catch!

rfcs/0003-declaration-engine.md
emilyaherbert: Expression is a better word.

rfcs/0003-declaration-engine.md
emilyaherbert: 👍

rfcs/0003-declaration-engine.md
emilyaherbert: Good call.

emilyaherbert:
There hasn't been much activity on this RFC page, largely because I have been working on this prototype: https://github.com/emilyaherbert/declaration-engine-and-collection-context-demo/tree/master/de. The prototype should 1) help explain the idea and 2) acts as a guiding light for our implementation. Trait constraints for generic functions are working in the prototype.

Also, I've basically reworked the entire RFC, so I'd love a second round of feedback. In the meantime, I will hack on a experimental branch in the Sway repo so that we can get things moving. Don't let this scare you away from iterating on the design though!

otrho (approved):

vaivaswatha (approved):

emilyaherbert:
ping @FuelLabs/sway-compiler would we be able to merge this?

mitchmindtree (approved):

sezna (approved):
🚀

PR #13, opened 2022-07-26
Intrinsics support in the Sway compiler.

vaivaswatha:
Markdown version: https://github.com/FuelLabs/sway-rfcs/blob/vaivaswatha/intrinsics/rfcs/0003-intrinsics.md

### DIFF
--- a/rfcs/0003-intrinsics.md
+++ b/rfcs/0003-intrinsics.md
@@ -74,11 +74,17 @@ To begin with, VM opcodes that are used through assembly wrappers in the core an
 
 A (non-exhaustive) list of intrinsics to support: `std::ops` (arithmetic, logical, comparison), `alloc`, `block::height`, `logging::log` etc.
 
+Because of the possibility of Sway supporting other targets in the future
+(such as EVM), VM specific intrinsics may be declared modularly, so that
+the toolchain can choose the ones to load based on the target.
+
 ## Testing
 For each core/std library function that we provide an intrinsic for, move the library function into the testsuite, and for every (prio) use of the library function in the testsuite, replace it with a call to the intrinsic and an assert that calls the (now in the testsuite) library function and asserts equivalence. For newly supported VM opcodes, add new tests
 
 ## Documentation
-TBD
+Given that Sway programmers are free to use intrinsics, it is essential
+that we document all intrinsics categorically in a book, perhaps
+alongside the standard library.
 
 # Drawbacks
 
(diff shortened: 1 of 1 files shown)


otrho:
Great.  This is a copy of the Google doc you made, so my comments remain the same, which is to say I agree with it. :slightly_smiling_face: 

It might be worth mentioning the possibility for Sway to target other backends (e.g., EVM) and so VM specific intrinsics (those which wrap low level instructions like `gas`) could/should be loaded/declared modularly.  The toolchain would get to choose which intrinsics are available based on the target.

And as mentioned, there's nothing stopping regular devs from using them so they'll need to be documented in a book, alongside the standard library I guess.

otrho (approved):
LGTM!

tritao (approved):
LGTM, all seems pretty reasonable.

mohammadfawaz (approved):
Nice and clear. Thanks!

sezna (approved):
I'm happy to see other backends mentioned. This is something we will be doing in the near future. Intrinsics allow us to move away from `asm` blocks, which essentially tie us to the FuelVM, and allow us to compile to alternate back ends. Commenting this here for posterity 😄

emilyaherbert (approved):
Looks great 👍  I agree with @sezna 's comment about other backends!

vaivaswatha:
> Per request in [FuelLabs/sway#2771 (comment)](https://github.com/FuelLabs/sway/pull/2771#discussion_r972933679), I think we could improve on the table based approach suggested here with a function-level attribute instead. Basically, intrinsic functions will lack a body `{ ... }` + get annotated with `#[intrinsic]`. Example:
> 
> ```rust
> #[intrinsic]
> fn my_magic_intrinsic<ty_params>(params);
> ```
> 
> This will live in `core`, which is the only "crate" that is allowed to use `#[intrinsic]`. The attribute tells the type checker's check-part that _"there is no body, and that's OK, because that is defined elsewhere"_. The type checker will however implicitly trust that the signature collected visavi parameter types and type parameters (+ constraints) are accurate, so calls to intrinsics are type checked like any other function calls.
> 
> Optionally, we could have a table-like situation where intrinsic _definitions_ (**not calls**) are type checked, e.g., like `rustc` does, https://doc.rust-lang.org/nightly/nightly-rustc/src/rustc_typeck/check/intrinsic.rs.html

This will require type constraints (for example, `__add` is a polymorphic function whose type parameter is constrained to be an integer) which isn't implemented yet, but will soon be. But I like this approach overall.

PR #15, opened 2022-08-26
Collection context

emilyaherbert:
[rendered](https://github.com/FuelLabs/sway-rfcs/blob/emilyaherbert/collection-context/rfcs/0005-collection-context.md)

[prototype](https://github.com/emilyaherbert/declaration-engine-and-collection-context-demo/tree/master/de_cc)


otrho (approved):

rfcs/0005-collection-context.md
```diff
@@ -0,0 +1,494 @@
...
+        return 1;
+    } else {
+        return alice_fn(n-1);
+    }
+}
+```
+
+When type checking `alice_fn`, the compiler is able to use the collection context to see that there does exist
+a `bob_fn` function, but that it is out of scope. This information is able to be used to gives better
+error messages and better error recovery.
```
JoshuaBatty, line 175: Nice! From the LSP side, we could provide a pop-up box asking the user if they would like to auto-import the missing library.

rfcs/0005-collection-context.md
```diff
@@ -0,0 +1,494 @@
...
+Relating to the `CollectionContext`, the Rust compiler uses
+[queries for demand-driven development](https://rustc-dev-guide.rust-lang.org/query.html). Admittedly,
+I'm not as familiar with this on a technical level as I would like to be, but my guess is that they are
+using a graph-based approach (combined with the Rust `DefId`) and to achieve the memoization and caching.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+I'd really like to get feedback on how the `CollectionContext` could be improved for use with the LSP.
```
JoshuaBatty, line 477: I can't really think of anything that could be improved at this stage. It all seems relatively straightforward.

I can see a situation in LSP where we have something like:

```
fn foo(bob: Bob) {
    let x = bob.
}
```

Could there be a way for the LSP to query the CC with the unfinished `bob` token to look up the closest edge in the graph? This way we could then go to that type and look up it's methods for autocompletion etc..

JoshuaBatty (approved):
Nice work @emilyaherbert! Thoroughly enjoyed reading this. Really excited to see this land.

sezna (approved):
Considering this is pretty much totally implemented, probably time for the RFC to go in 😄

mohammadfawaz (approved):
Gave my verbal approval to Emily yesterday. Making it official 😀

PR #18, opened 2022-10-03
Fix broken links

sezna:


mohammadfawaz (approved):

nfurfaro (approved):
lgtm.

Braqzen (approved):
Why are these blobs instead of "regular" links?

sezna:
> Why are these blobs instead of "regular" links?

`/blob/master` and `/latest` don't pin to a specific rev, they pin to the latest one. If a file is renamed, the link becomes a 404. This is what happened to a couple of the links, causing the CI to fail. Using a specific rev prevents the content behind the link from changing.

AlicanC (approved):

PR #19, opened 2022-10-06
Configuration Time Constants

sezna:
[rendered](https://github.com/FuelLabs/sway-rfcs/blob/a5c69ccf4b44fbcdd14a14f6af9f50da8b02eb3d/rfcs/0006-configuration-constants.md)

You may notice that config time constants were already added to the language (#2549). This implementation did not actually address the use cases it was intended for, though, and so this is a request for comments on v2 of the notion. The new implementation of configuration time constants will be more complicated as it involves interactions between the SDK and Sway via a new descriptor file, but it does maintain independence between the two tools so updating constants won't require a recompile.


rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,99 @@
...
+- RFC PR: [FuelLabs/sway-rfcs#0006](https://github.com/FuelLabs/sway-rfcs/pull/19)
+- Sway Issue: [FueLabs/sway#1498](https://github.com/FuelLabs/sway/issues/1498)
+
+# Summary
+
+[summary]: #summary
+
+Configuration time constants can be conceptualized as traditional environment variables. Some bytecode has been compiled by the Sway compiler, and the SDK would like to configure some behavior of that bytecode with additional inputs that won't trigger a recompile.
+
+There is a similar feature that was implemented [here](https://github.com/FuelLabs/sway/pull/2549), but that was a mistaken interpretation of the requirements. In #2549, a recompile still occurs and the new values are injected via `Forc.toml`, when we actually want these values to be injectable by the SDK.
```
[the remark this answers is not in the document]
iqdecay: Would the link still work in the `md` file?

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,99 @@
...
+
+This supports the use case of configurable values post-compile-time but pre-runtime, similar to Solidity's `immutable` concept or environment variables in traditional programming. This allows for, e.g., updating contract addresses without recompiling the bytecode.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+When a Sway program is compiled, a set of artifacts are produced. Included in these artifacts are a JSON ABI descriptor, a configuration-time-constant descriptor, and the actual bytecode. The object we are discussing here is the configuration-time-constant descriptor, which describes the specific offset within the bytecode, in terms of bytes, to the data section entry for a particular configuration variable.
+
+The configuration variables for a program can be identified by the `configurable` keyword within the Sway language. This denotes a constant within the program that has no known value at compile time. The value will be provided at configuration time by the SDK or some other consumer of the configuration-time-constant descriptor. As an example:
```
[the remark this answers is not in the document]
iqdecay: fwiw I also think `configurable` is more user-friendly

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+One major drawback is the additional cognitive complexity this adds to the compilation process. If any bug is introduced or the overwriting of the bytecode goes awry in any way, it will result in confusing and inconsistent undefined behavior. We will need to introduce sufficient checks to ensure that an end user would never encounter this situation.
+
+This increases our reliance on the correctness of the ABI encoder in the SDK, and any version mismatches or bugs in the encoder will result in similarly undefined and unpredictable behavior.
+
+This change also further couples the language to the SDK, further minimizing the use case of writing Sway without the SDK. This can be alleviated by introducing mechanisms to `forc` for defining these values manually, as mentioned in the guide-level explanation.
```
iqdecay, line 70: This also potentially means that we are going back to the old issue of circular deps between `fuels-rs` and `sway`?

sezna: Since we would rely  on the abi decoder/encoder for this, we hopefully would avoid having to import `sway`. In fact, it is a goal of this effort to avoid having to import `sway`, so if it isn't possible, we will have to refactor.

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+One alternative is to introduce an environment variables/configuration-time constants section to the transaction format. This could be cleaner, but the impact on Sway would be minimal. Sway would still output a descriptor file, introduce the `configurable` keyword syntax, etc., and the majority of this RFC would still apply. The benefit of adding this to the transaction format would be potential simplification of the process. A couple negatives would be additional churn on the client, VM, and SDK; and further coupling of specifically Sway to the FuelVM.
```
vaivaswatha, line 76: Another alternative would be to get `configurable` values from the VM via a new operation.

Like now (i.e., as in this RFC), the compiler would, in the descriptor file, describe the `configurable`, and in addition provide an integer ID to it. This integer ID would be the argument to the new VM op, and the VM would return the value. The downside to this would be that we have an OP whose return type is not fixed. At the compiler level this is okay because the compiler knows the return type (based on the ID passed) and do its type-checking job.

vaivaswatha (approved):

rfcs/0006-configuration-constants.md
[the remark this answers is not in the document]
sezna: I personally do like `configurable`, because it reads like the value is a configurable variable. `config` sounds stateful to me, like a config file or something. Happy to bikeshed though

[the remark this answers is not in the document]
sezna: This is a lot more syntax, but the `extern` block does do a nice job of highlighting the externality of the configuration constants.

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,99 @@
...
+
+The change is not breaking as it is entirely new functionality.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+One major drawback is the additional cognitive complexity this adds to the compilation process. If any bug is introduced or the overwriting of the bytecode goes awry in any way, it will result in confusing and inconsistent undefined behavior. We will need to introduce sufficient checks to ensure that an end user would never encounter this situation.
+
+This increases our reliance on the correctness of the ABI encoder in the SDK, and any version mismatches or bugs in the encoder will result in similarly undefined and unpredictable behavior.
```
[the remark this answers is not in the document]
sezna: Yeah, it is impossible to avoid at some point, but worth mentioning.

rfcs/0006-configuration-constants.md
sezna: This is a very interesting idea. However, that sounds like runtime environment variables. Here, our objective is to introduce variables after compile time but _before runtime_. The deployed bytecode should not have any variability defined at runtime -- the constants are provided by the SDK during or before the deploy.

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,101 @@
...
+
+This change also further couples the language to the SDK, further minimizing the use case of writing Sway without the SDK. This can be alleviated by introducing mechanisms to `forc` for defining these values manually, as mentioned in the guide-level explanation.
```
digorithm, line 72: > This can be alleviated by introducing mechanisms to `forc` for defining these values manually, as mentioned in the guide-level explanation.

Wouldn't `forc` still use the SDK's functionalities to encode and insert the encoded values in the bytecode? Otherwise, we would have this (critical) functionality implemented twice in different places (`forc` and in the SDK).

So we either duplicate these implementations, or we _don't_ alleviate the drawback mentioned above because it would still depend on the SDK?

digorithm (approved):
This is looking good to me. There are a few things I couldn't find/think-of answers to. First is agreeing on how this will work for deployments _without_ the SDK; are we settling on flags passed to `forc deploy` and then `forc` using the SDK's functionalities to encode these values onto the bytecode?

Second, but very minor, is what the errors from omitting these `configurable`s or passing malformed values will look like. I think for malformed values, the SDK will provide proper error messages, so it should work out of the box 🤔. 

Just some thoughts. Overall, the plan looks good to me.

mohammadfawaz (commented):
Looks good to me! Thanks 😄 

Some things that are worth mentioning:
1. Configuration time constants are not allowed to be optimized away or constant folded or really touched in any way (for example, a config struct cannot be broken into into its individual elements). They _always_ need an actual spot in the bytecode. So, the optimizor has take that into account.
2. Are there any restriction on where configuration time constants can be declared? Are they only allowed at global scope for example? I don't recall what we do with `const` today.
3. Can you add a note about the types allowed? If we want to allow any type (not just primitive ones), then we should say that the SDK needs to follow the same data layout described in the ABI spec when injecting the constants in the bytecode.
4. I'm assuming the `pub` keyword can be used, just like with regular constants?
5. I think the RFC should specify how these constants are initialized. I'm assuming they will be initialized to zero (and whatever that means for enums etc.) but we should be explicit about that. That being said, do also want to allow explicit initialization?

rfcs/0006-configuration-constants.md
sezna: With `forc`, in order to avoid importing the SDK into `forc`, we could pass the values into the build config. Since we have direct access to the compiler and its memory layout when building via `forc`, we could use that evaluator directly. Ideally, these two things should never differ since the encoder should match the same layout. Do you think there are other advantages to using the ABI encoder?

kayagokalp:
> I'm assuming the pub keyword can be used, just like with regular constants?

Just to provide some input about this, while working on https://github.com/FuelLabs/sway/pull/3016, I realized that having an option to have `pub` config time constants would help me a lot. I am injecting the contract ids of the contracts that are declared under `[contract-dependencies]` section of the `Forc.toml` automatically so that they can be used roughly like 

```rust
let contract_b_address = contract_b::CONTRACT_ID;
```

in `contract_a`, which declares `contract_b` under `[contract-dependencies]` (basically a deployment dependency).

To do so, I wanted to utilize config time constants and insert the submodule created for `contract_b` with the `CONTRACT_ID` constant that holds the address for `contract_b` to the namespace of `contract_a`. Since I cannot declare a config time constant to be public rn, i am unable to use the injected `CONTRACT_ID` of `contract_b` from `contract_a`.

emilyaherbert (changes requested):
Just one question.

rfcs/0006-configuration-constants.md
digorithm: Ah, I completely missed the idea of using the compiler itself to do it; that means you don't need a _third_ implementation, and that's what I was concerned about!

mohammadfawaz:
Regarding the JSON spec, this is what I propose: https://github.com/FuelLabs/fuel-specs/pull/419/files (just a generalization of what you already have).

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,125 @@
...
+
+There is a similar feature that was implemented [here][pr_2549], but that was a mistaken interpretation of the requirements. In [#2549][pr_2549], a recompile still occurs and the new values are injected via `Forc.toml`, when we actually want these values to be injectable by the SDK.
+
+[pr_2549]: https://github.com/FuelLabs/sway/pull/2549
+
+# Motivation
+
+[motivation]: #motivation
+
+This supports the use case of configurable values post-compile-time but pre-runtime, similar to Solidity's `immutable` concept or environment variables in traditional programming. This allows for, e.g., updating contract addresses without recompiling the bytecode.
```
mitchmindtree, line 20: It would be great to have some more fleshed out motivations here, otherwise it's tricky to weigh up the benefits vs the added complexity.

Currently, it looks like the only listed motivating example would already be addressed by https://github.com/FuelLabs/sway/pull/3016 in a manner that doesn't require a "configuration-time" step or new language features at all.

mitchmindtree (commented):
@kayagokalp we already discussed this briefly in tooling sync, but just replying here too for readers of this discussion:

In https://github.com/FuelLabs/sway/pull/3016 we can use the same approach as used by the *old* config time constant implementation (i.e. manually adding the `CONTRACT_ID`s to the relevant dependency namespace in `forc-pkg`). Otherwise, that PR should remain unrelated to this proposal and the idea of configuration time constants in general.

rfcs/0006-configuration-constants.md
sezna: I think the only motivations are avoiding recompiling from the SDK. @digorithm am I forgetting something?

mitchmindtree: If our only motivations are avoiding recompilation, my next question would be: Why?

Is it just the compilation time that is bothersome? If so, I think it's worth noting that we haven't yet had a chance to assign any serious resources to bench-marking or optimizing compilation times beyond taking a quick look on the rare occasion they become unbearable. Perhaps it's worth considering investing the resources that would otherwise be involved in implementing/maintaining this new feature into optimizing compilation times?

I think your section on *Interactions with optimization passes* is an important one too, and I worry a little that we might end up accidentally blocking or over-complicating some future optimisation work unnecessarily.

In general, this feature seems big, complicated and annoying enough to maintain that it should probably have some strong motivations - I think we might want stronger motivations than we currently have?

@digorithm if you (or anyone else) have any more thorough motivating examples or workflows that would be great!

mitchmindtree:
Another subtle implication of this feature would be that all configuration time constants would need to be known in order to produce a contract's ID.

This would complicate `forc`'s ability to automatically forward contract IDs during multi-contract builds or deployment as done in https://github.com/FuelLabs/sway/pull/3016. Every entry under `[contract-dependencies]` would need to specify each configuration time constant value too.

rfcs/0006-configuration-constants.md
```diff
@@ -46,7 +49,7 @@ fn main() { ... }
 }
 ```
 
-If you are not using the SDK, it will be possible to define these values either by passing a flag to `forc` or including them in the `Forc.toml` manifest file. If both are present, then the flag takes priority over the manifest.
+If you are not using the SDK, it will be possible to define these values by specifying a `constants.sw` flag. This will work with the [notion of contract dependencies](https://github.com/FuelLabs/sway-rfcs/pull/19#issuecomment-1286351753). This Sway file must export the required constants as specified by the `configurable` block.
```
mitchmindtree, line 52: The ability to specify these values in `Forc.toml` is more about the requirement for `forc`'s "Contract Dependencies" feature - there's nothing stopping the user from altering these values with the SDK afterwards I guess. Maybe this could be something like:

```suggestion
As `configurable` values will affect a contract's ID, `forc` users must be able to specify these values for each of their contract dependencies. To enable this, `forc` will allow users to specify a `config/<contract-dependency-name>.sw` file for each contract dependency that contains configurable values. This Sway file must export the required constants as specified by the contract dependency's `configurable` block. These files will be compiled independently for each contract dependency, and the resulting `configurable` bytecode section for each dependency will be used to replace its default `configurable` bytecode section.
```

mitchmindtree (commented):
One last suggestion, otherwise looking good!

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,128 @@
...
+
+The configuration variables for a program can be identified by the `configurable` keyword within the Sway language. This denotes a constant within the program that has no known value at compile time. The value will be provided at configuration time by the SDK or some other consumer of the configuration-time-constant descriptor. As an example:
+
+```rust
+// main.sw
+script;
+
+configurable {
+    CONTRACT_ADDRESS: b256,
+    PRICE_RATIO: u64,
```
mohammadfawaz, line 36: We should either support initializers here or be clear in the text that these are initialized to 0, whatever 0 means for each possible types (just 0 bytes essentially).

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,128 @@
...
+
+No two configurables in a program should have the same name to prevent collisions.
+
+The range from the offset to the offset plus the size of the type will represent bytes within the bytecode that will be overwritten by the SDK, which will write values according to the ABI encoder's memory layout, which is consistent with Sway's memory layout.
+
+The change is not breaking as it is entirely new functionality.
+
+Configuration time constants may not be defined in any library code and must be defined at the top level. Should a library need to reference a configurable value, it should utilize design patterns such as the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern) to ingest the constants from the top level.
+
+The `pub` keyword will operate as normal: if it is `pub`, then it is public and importable from elsewhere in the Sway program. If not, then it is only referencable in the scope it is defined within.
```
mohammadfawaz, line 69: So these two are a big conflicting now since `pub` is not really relevant outside a library for the time being. Besides, with the new syntax, `pub` does not really have a place?

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Deploying with and without the SDK
+
+With the SDK, the deployment and assignment of values will all be handled within the SDK's API. Without the SDK, `forc` will accept command line arguments to specify these values, or allow for their specification in the manifest. In the future, we may wish to introduce more methods of defining these values, like environment variables, but it will not be necessary for the MVP. If the values are specified in more than one way, the SDK would take precedence. This is because the SDK does not actually compile the bytecode in this process; it merely overwrites the indices. This is final and precedence cannot be specified by the user.
```
mohammadfawaz, line 73: > Without the SDK, `forc` will accept command line arguments to specify these values, or allow for their specification in the manifest.

That's not how we're going to do it anymore, right? We're going to go with `constants.sw` approach?

rfcs/0006-configuration-constants.md
```diff
@@ -0,0 +1,128 @@
...
+
+## Failure to provide constants
+
+If the constants are not provided, the SDK and/or `forc` should throw an error. Both tools have knowledge of the required constants due to the descriptor file, so they may check for the constants' presence and report a descriptive error if any are missing.
```
mohammadfawaz, line 77: By `forc` here we mean `forc deploy` specifically? I assume `forc build` shouldn't care as it should be able to emit bytecode with default initialized config constants, that are later specified by the SDK.

rfcs/0006-configuration-constants.md
mitchmindtree: Good catch! Agreed, I don't think we should make any promises about offering command line arguments in `forc` here. I made a suggestion [here](https://github.com/FuelLabs/sway-rfcs/pull/19#discussion_r1003867760) that clarifies why it's needed for `forc` and the proposed behaviour.

rfcs/0006-configuration-constants.md
mitchmindtree: `forc build` *does* care due to [the Contract Dependencies feature](https://fuellabs.github.io/sway/master/forc/manifest_reference.html#the-contract-dependencies-section). During `build`, we generate contract ID constants for each contract dependency, and the configurable value for each contract dependency must be known in order to produce the correct ID.

mitchmindtree: To elaborate, this doesn't stop the user from updating `configurable` values later on using the SDK to create a contract factory or whatever.

It's just that we should expect that users **will** misuse the feature and expose configurable values from contracts that don't really need them to be configurable. As a result, we need to be able to specify these values when these contracts are specified within a forc project's `[contract-dependencies]` table, otherwise we can't guarantee a correct ID for those users who actually want a declarative approach to contract deployment.

---

W.r.t. this "Failure to provide constants" section, @mohammadfawaz I like your idea [here](https://github.com/FuelLabs/sway-rfcs/pull/19#discussion_r1003901756) of requiring that any declared `configurable` values always have some initial value (whether explicitly specified or defining some "zeroed" default for each type). That would remove need to throw an error at all I guess, as at the very least the configurable bytecode should always be valid?

nfurfaro:
Looking good to me, and seems like any questions I had have been asked and addressed already.

mitchmindtree (approved):

mohammadfawaz (approved):

kayagokalp (approved):

nfurfaro (approved):

nfurfaro:
@dmihal Contract factories will have to be done offchain. It will be possible to calculate the Id of any given contract + the specific configurable constants for that contract using the SDK, and compare it with the deployed contract to verify its correctness.

mohammadfawaz:
@emilyaherbert requesting your review on this RFC since you requested changes.

JoshuaBatty (approved):

JoshuaBatty:
pinging @emilyaherbert for approval.

PR #22, opened 2023-01-10
Support the EVM as a backend target.

otrho:
Authored by @jtriley-eth.  [Here](https://github.com/FuelLabs/sway-rfcs/blob/evm/rfcs/0007-evm-support.md)

Here's a proposal for supporting the EVM.  It still asks a few questions rather than providing a concrete path forward but that's what RFC PRs are for.

### DIFF
--- a/rfcs/0007-evm-support.md
+++ b/rfcs/0007-evm-support.md
@@ -669,3 +669,73 @@ non-existent. This will allow EVM engineers to design more secure smart contract
 alike, will boost Sway's total TVL in the [Defi LLama Language TVL Listings](https://defillama.com/languages),
 and will spark interest in the language enough to bring engineers into the Fuel ecosystem with
 marginal learning overhead.
+
+## ERC Compliant Casing Option
+
+A nice-to-have specifically for the EVM would be to have a way to modify the function signature and
+selector to conform to "Etheruem Request for Comment" standards. Sway follows patterns of Rust,
+which, according to the [Rust Style Guide](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html),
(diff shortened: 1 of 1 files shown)


JoshuaBatty (approved):

rfcs/0007-evm-support.md
```diff
@@ -0,0 +1,741 @@
...
+[project]
+authors = ["user"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "example"
+target = "eofv1"
+```
+
+Development, Testing, and Deployment can be done through Hardhat or Foundry using a simple plugin
+for each.
```
kayagokalp, line 117: Hmm this makes me think if we are planning to support sway unit tests for EVM backend cc @mitchmindtree 

Currently `forc-test` is coupled with `FuelVM`, if we are going to be supporting tests for EVM as well we might need to think about it a little.

kayagokalp (approved):

rfcs/0007-evm-support.md
mitchmindtree: Yeah I think to start we'd want to find a nice embeddable EVM interpreter - I'm guessing there might be a lib from foundry we could plug-in for this?

otrho: We already have an EVM testing framework set up for the E2Es which might work for unit tests too.  Not sure, @tritao looked at it initially.

kayagokalp: Nice 👌, it will be a good reference. Thanks!

SwayStar123 (approved):
Ship it

Braqzen:
> Ship it

Box it and transport it

PR #23, opened 2023-02-11
Introducing `StorageKey`

mohammadfawaz:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/05b7594b0b64ad91baaedb732994fc3f251b0042/rfcs/0008-storage-handler.md)
 
Related sway issue: https://github.com/FuelLabs/sway/issues/3043

Relevant issues:

### Related Issues
- [ ] https://github.com/FuelLabs/sway/issues/3202
- [ ] https://github.com/FuelLabs/sway/issues/2465
- [ ] https://github.com/FuelLabs/sway/issues/2585
- [ ] https://github.com/FuelLabs/sway/issues/2639
- [ ] https://github.com/FuelLabs/sway/issues/1182
- [ ] https://github.com/FuelLabs/sway/issues/3419


rfcs/0008-storage-ref.md
```diff
@@ -0,0 +1,249 @@
...
+This should still allow writing the following:
+
+```rust
+storage {
+    my_map: StorageMap<u64, u64> = StorageMap {}
+}
+
+my_map.insert(0, 0);
+let x = my_map.get(0);
+my_map.as_storage_ref().remove(0);
```
mitchmindtree, line 145: Not sure I follow this line - is the `as_storage_ref()` necessary/intended here?

rfcs/0008-storage-ref.md
```diff
@@ -0,0 +1,249 @@
...
+}
+
+storage {
+    s: MyStruct, 
+}
+
+fn foo() {
+    let x_ref: StorageRef<u64> = storage.s.x.as_storage_ref();
+    let y_ref: StorageRef<b56> = storage.s.y.as_storage_ref();
+    let s_ref: StorageRef<MyStruct> = storage.s.as_storage_ref();
```
mitchmindtree, line 69: Does a storage access have a valid type at all *without* having called `as_storage_ref()`?

If not, could we perhaps assume that any field access on storage (i.e. `storage.x.y.z`) is always a `StorageRef<T>`?

Hmm I guess this might require some compiler magic to enable the access of the field `y` on the `StorageRef<T>` resulting from `storage.x` and so on...

rfcs/0008-storage-ref.md
```diff
@@ -0,0 +1,249 @@
...
+}
+```
+
+In order to use a storage reference, one can call `store` and `get` as follows:
+```rust
+let x_ref: StorageRef<u64> = storage.s.x.as_storage_ref();
+x_ref.store(42); // equivalent to `storage.x = 42`
+let x = x_ref.get(); // equivalent to `let x = storage.s.x`
```
mitchmindtree, line 77: This `get()` isn't really equivalent as the comments suggest, as `get()` in the proposed API returns `Option<T>`.

rfcs/0008-storage-ref.md
```diff
@@ -0,0 +1,249 @@
...
+    key: b256,
+}
+```
+
+We can implement `store` and `get` as follows:
+
+```rust
+impl<T> StorageRef<T> {
+    #[storage(read)]
+    pub fn get(self) -> Option<T> {
```
mitchmindtree, line 35: I can imagine getting frustrated working around this `Option` for references that you *know* must be valid, e.g. `storage.foo` in `storage { foo: u32 = 42, }`. Was the intent for this to be `pub fn get(self) -> T`?

I imagine there will be cases where you *would* want to return an `Option`, e.g. if you ever have a reference to an element of a storage *collection*, as it's possible the element may get removed between when you construct the reference and when `get` is called as we have no borrowcheck in Sway. Maybe there needs to be different types for the two cases? I.e. 1 where you know the element will always be valid and 2. another where the element may have been freed? :thinking:

rfcs/0008-storage-ref.md
mohammadfawaz: So I'd like methods of `StorageMap` to work with and without the `as_storage_ref`, which is why I'm suggesting https://github.com/FuelLabs/sway/issues/4047.

rfcs/0008-storage-ref.md
mohammadfawaz: The type of `storage.x.y.z` is just the same type as `z`. The problem with this is that calling `as_storage_ref` will only be valid if called directly using `storage.x.y.z.as_storage_ref()` but not valid if called as below:
```
let z = storage.x.y.z;
let z_ref = z.as_storage_ref()`
```
despite `z` having the same type as `storage.x.y.z`.

Another reason why making `storage.x.y.z` of type `StorageRef` might be a cleaner approach, though I'm reluctant to make this change because it complicates the syntax for getting and updating simple storage variables.

rfcs/0008-storage-ref.md
mohammadfawaz: That's true.. this is an oversight as `get` was changed recently to return an `Option`.

rfcs/0008-storage-ref.md
mohammadfawaz: This is pretty annoying yes - when I started thinking about this, `get` did not return an option but now it does. What do you mean by two different types? Two different `StorageRef` types?

rfcs/0008-storage-ref.md
mitchmindtree: > The type of storage.x.y.z is just the same type as z.

I realize this is the case today, but are we sure we want to continue to allow this?

E.g. if the field  `z` has the type `Z` and `Z` has some method `foo` that takes `ref mut self`, what happens when you do `storage.x.y.z.foo()`? Does it update storage at the field `z`? The method implementation for the type `Z` would behave differently when the type is used in storage, vs when the method is called on an instance of `Z` that is *not* in storage right? Could the fact that `Z` may have a different layout (whether used in the context of storage or not) not have an impact on the behaviour of the method?

It seems like, for any storage access, we might want to *require* that it's through the lens of some sort of storage reference?

mitchmindtree:
*Sharing some of what was discussed in the lang call.*

## Aliasing and Use-After-Free

Something to consider is that this is the first time in Sway that we're able to *return* a reference to something. This means we've entered the lands of potential use-after-free (or use-after-unexpectedly-modified-in-any-way).

Here's an example of some new highly-questionable yet valid Sway code:

```rust
let elem_ref = storage.foo_map.get(elem_key);
storage.foo_map.remove(elem_key);
elem_ref.do_something(); // uh oh
```

Here's another:

```rust
// storage.owner_groups is a `StorageMap<StorageVec<Address>>`.
let bobs_friends = storage.owner_groups.get(group_key);
// some code...
storage.owners.insert(group_key, alices_friends);
// some more code....
for bob_friend in bobs_friends {
    bobs_friend.send_asset(asset); // uh oh... why does alice's friends have our money?
}
```

These are the kinds of cases that Rust's borrow-checker will prevent at compile-time with true "references".

## Tackle a `&'a T` or `Rc<T>` *type* first?

I think the trickiest thing about designing for storage right now is having no standard safe way of returning actual, safe *references* in Sway in general. The more I try to come up with alternatives, the more I get cut short at having no way to write a method that returns a reference in a safe way :woozy_face:  It's particularly relevant for storage as access is expensive, but it's probably something we will want as a part of designing APIs in general.

Should we actually consider investigating a more general borrow-checked or ref-counted reference before trying to tackle "storage" references in particular?

mohammadfawaz:
@mitchmindtree I thought I’d update the RFC to have the expression `storage.x.y…` return a `StorageHandler` (I renamed `StorageRef` to `StorageHandler`) to see how things would look like.. I think this simplifies things a lot of things. The use-after-free issue is still there, but I wonder if the fact that `get` returns an `Option` is sufficiently safe? The VM is able to tell us at any given point whether a storage slot is valid or not. This is of course a runtime check unlike the borrow checker, but maybe that’s a sufficient first step?

As for ergonomics, the user now has to write `let x = storage.x.get().unwarp_or(..)` or something like that, which is not very ergonomic sadly, but maybe we can add another method here `get_or_default()` so that this looks like `let x = storage.x.get_or_default()` which is _technically_ is the current behavior (edited) 

The fact that storage is more like external I/O than regular memory, means a lot of the checks need to be at runtime. We can’t guarantee much at compile time. Imagine a situation where one contract methods unsets or clears a storage slot, and then another method tries to read it.. there is no way for the compiler to know when this is safe because it doesn’t know the order in which these methods will be called. So, despite `storage.x.get()` returning an `Option` not being the most ergonomic, it seems like the most safe to me?

Of course, one alternative here is to go back to the way things were before the VM started exposing the set/unset flag and just assume zero bytes when a storage slot that hasn’t been written to is read from.

otrho:
> The fact that storage is more like external I/O (global) than regular memory, means a lot of the checks need to be at runtime

Yep, thinking of them as references doesn't really make sense as all I/O is transactional, so use-after-free isn't really a thing.  We could instead have read-only or write-only storage handlers to protect against unwanted out-of-scope modifications, much like a traditional file handle, which is what I keep coming back to mentally.  I guess the whole point of storage ref is to allow writing to storage in libraries -- if you wanted to provide read only access then you'd probably just read it out and pass a 'copy'.

Storage is more like a database, with typed fields and an API, rather than just global memory.  We should think of safety in terms of how a database would protect its data.

mitchmindtree:
I agree that the I/O model sounds like a much better suit. I'd love to see an exploration of a more transactional API as you mention @otrho. The more we work on a storage design, the more I think the suspect the current representation just has too many edge cases for anything that that does some kind of dynamic allocation, or for structs that happen to have multiple fields, etc. It looks and feels a lot like a struct, but doesn't really behave like one at all under the hood. We could go down the road of keeping up this deception like solidity does, but surely that will just continue to complicate our internals pretty severely?

> I think get() should definitely return an Option.

One thing worth keeping in mind is that for fixed-size data, we *could* technically guarantee that certain keys could always have valid data, but this would require not publicly exposing the `store` intrinsic (or restricting its use to something like an `unsafe` block).

Is it maybe worth distinguishing between *statically*-allocated storage and *dynamically*-allocated storage in the language? E.g. we could reserve the use of the `storage { .. }` syntax for declaring fixed-size, statically allocated values that are always known to be valid, and then provide a separate, more file-like abstraction over the rest of storage on which we can build our more dynamic abstractions like `StorageMap` and `StorageVec`?

I guess this might just complicate the API unnecessarily to save on what is essentially a single bounds-check, which we might be able to optimize away in the future if we get smarter about detecting fixed-size storage elements.

Maybe it's worth mocking up what a I/O style interaction with storage might look like? Then maybe once we have an honest representation of what storage looks like in the language, we can start to imagine what the higher-level ergonomics around it might look like?

mohammadfawaz:
> Maybe it's worth mocking up what a I/O style interaction with storage might look like? Then maybe once we have an honest representation of what storage looks like in the language, we can start to imagine what the higher-level ergonomics around it might look like?

I've added the counter example in the RFC using the approach of `storage.x` being a `StorageHandler` itself and how `.get` and `.store` would look like. Maybe `storage.x` should return something else that's known to be valid and we can just call `unwrap` in the definition of `StorageHandler::get`? As you said, the only way to invalidate a slot is by calling `clear` which we can hide.

Also, when calling `get` on a `StorageMap` today, we get back an `Option` wrapping the data because the data might have been removed from the map. We don't return a reference or a handler to the data. So the "dereferencing" part is happening internally. The RFC has some examples of that too.

Really a `StorageHandler` only _exists_ for storage variables defined in a `storage` block and have exact state indices and potentially field indices if the storage variable is a struct and we're accessing a field in it.

I'm starting to feel better about what we currently have here and I wonder if this is a good middle ground? I can add some more examples and we can evaluate.

rfcs/0008-storage-ref.md
mohammadfawaz: No longer relevant since `storage.x.y.z` now returns the storage handler itself.

rfcs/0008-storage-ref.md
mohammadfawaz: I added this as a drawback and we should discuss how to make this more ergonomic.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+# Motivation
+
+[motivation]: #motivation
+
+The current approach for reading and writing storage slots/variables has multiple flaws, despite being quite ergonomic:
+
+1. There is currently no way to reason about the location of a particular storage variable, which makes passing "references" to storage variables impossible. One workaround is manipulating and reading storage slots manually via `std::storage::store` and `std::storage::get`, but this approach is quite unsafe. Alternatively, one could declare a `trait` in a library and implement it in a contract and have its methods access storage variables. Those methods are then used in the library to access storage indirectly. This does not solve the full problem and feels more like a workaround than a real solution.
+1. Dynamic storage types, such as `StorageMap` and `StorageVec`, currently require a "hacky" compiler intrinsic called `__get_storage_key` which is not very well defined and is hard to use. Introducing the concept of a `StorageHandler` will helps us completely remove this intrinsic.
```
nfurfaro, line 19: 2.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+1. The current implementation of dynamic storage types prevents th-em from being used as struct fields or as type parameters of other storage types (e.g. `StorageVec<StorageVec<u64>>`).
```
nfurfaro, line 20: 3.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+    s: MyStruct = MyStruct { x: 0, y: ZERO_B256 }, 
+}
+
+fn foo() {
+    // Get a storage handler to the struct field `x` directly and then use it to store `42` in `x`.
+    let x_handler: StorageHandler<u64> = storage.s.x;
+    x_handler.store(42);
+   
+    // Call `store` and `get` directly on `storage.s.y`
+    storage.s.y.store(ZERO_B256);
```
nfurfaro, line 74: Is there any benefit to allowing users to call `store` on `storage.s.y` directly here without using a `StorageHandler`?
It might be simpler to just allow the one way, i.e: 
```
let y_handler: StorageHandler<u64> = storage.s.y;
    y_handler.store(ZERO_B256);
```

mohammadfawaz: This really is just a byproduct of the fact that `storage.x.y` is an expression and so calling `.store` directly or through a temporary variable is intrinsically supported by the language, so no extra work there :)

rfcs/0008-storage-handler.md
nfurfaro: All 3 points above are labeled `1.` :)

otrho: That's how Markdown works. :)  If you look at the rendered version they are sequential.  Makes it easier to edit the list, to insert new items later on.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+> **Note**
+> Struct variables are laid out in the same way today.
+
+For example, the storage variable `str` above is a string containing `99` characters and its size is `99` bytes which span 2 storage slots. It is also assigned index `2`. Therefore, `str` spans slots with keys `sha256("storage_2")` and `sha256("storage_2") + 1`.
+
+Variable `s` on the other hand, is a nested struct and is assigned index `3`. Therefore, each of its fields (and subfields) is assigned a separate key starting with `sha256("storage_3")`:
+
+- `s.a` is assigned `sha256("storage_3")`.
+- `s.b.c` is assigned `sha256("storage_3") + 1`.
+- `s.b.d` is assigned `sha256("storage_3") + 2`.
```
otrho, line 185: OK, a couple of questions here.  Firstly, won't a 99 byte string require 4 32-byte slots?  99 / 32 = 3.09

I don't think the each-struct-field-goes-in-its-own-slot thing will fly.  If I have a struct with 4 ints in storage and I want to read and use them all, I am going to want them to be a single storage read and not to waste gas on 4 separate reads.  Isn't this why the Eth devs pack everything so tightly all the time?

If anything, putting each field into its own slot can only ever be _less_ efficent assuming that storage reads and writes are more expensive than a shift and a mask.  So I guess that is the question.

otrho (commented):
I am liking this approach.  Transactional access leaves no ambiguity about the nature of storage and it's simple.

I'm thinking if we do insist on `get()` returning `Option` and we keep `clear()` then we could also allow storage variables to not be initialised in the `storage` decl block.  Without the storage-is-special-memory model then the initialiser isn't required.  This is also simpler.

rfcs/0008-storage-handler.md
otrho: And in the example at the top you show ` let s:Option<MyStruct> = s_handler.get();` which by the splitting logic would actually be two storage reads?

Equivalent to:
```rust
storage.s.x.get().and_then(|x| (x, storage.s.y.get())).and_then(|(x, y)| struct MyStruct { x, y })
```

mohammadfawaz: >  Firstly, won't a 99 byte string require 4 32-byte slots? 99 / 32 = 3.09

Lol yes of course

As for the struct splitting: 
1. This is how we do it today so this is not a regression. Not to say this justifies doing it that way.
2. One issue with packing is that if I want to update one internet field in a struct, I may need to read the whole slot where the field lives, update the 8 bytes where the integer is stored, and then store the whole thing back (read-modify-write). There is also the risk that some fields may now span 2 slots (e.g.: for (u64, b256), the b256 will spill over to the next slot, so updating it requires two read-modify-writes). In any case, the idea is that accessing or updating a single field is more efficient we the approach described here, but picking is more efficient when accessing or updating the whole struct.

otrho: Ahh, right, read-modify-write is a thing.  We might need to offer both packed and unpacked options in the future, depending on what enough of the devs end up wanting.

So every leaf in a collection of nested structs gets its own slot?  And then they all get sequential keys based on declaration order?  If so we need to make sure that order is absolutely preserved thoughout the AST and beyond.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+Sway code using storage maps and vectors will not look any different than before after this RFC is implemented.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+There are two main drawbacks for the approach proposed above:
+
+1. Reading and writing a storage variable is now a bit more complicated and less ergonomic than before because we now have to call `store` and `get` manually. We also have to handle the `Option` returned by `get` which can be quite verbose. We should consider making this more ergonomic by introducing a simpler syntax that de-sugars to the API calls above. For example, we could de-sugar something like `storage.x = 42` to `storage.x.store(42)`. We could also consider making `get` return the data directly instead of an `Option` if we completely disallow storage slots corresponding to storage variables from being invalidated (For example, remove `std::storage::clear`).
```
mitchmindtree, line 304: While considering these alternatives, we should think about whether or not we really *want* to provide simpler syntax.

I think the user should at least feel the difference in cost of using storage. Hiding interaction behind e.g. an assignment operator, while *slightly* more concise, looks and feels too much like a regular in-memory stack assignment, and I worry new users will confuse the costs of the two?

mitchmindtree: > We could also consider making `get` return the data directly instead of an `Option` if we completely disallow storage slots corresponding to storage variables from being invalidated (For example, remove `std::storage::clear`).

What is the current use-case for `clear`?

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+2. The second drawback is related to the fact that structs take up more storage slots than they actually need. This behavior is **not** new. There is a tradeoff here to be considered. If we make structs tightly packed, then manipulating a single field becomes more expensive (potential read-modify-write patterns or fields "spilling" over to the next slot), but the number of storage slots required may become lower. We might want to consider a `#[packed]` attribute in the future to let the user decide on one way v.s. the other. Another thing to consider with tightly-packed structs is that the type `StorageHandler` would then also require an additional field that points to a particular location _inside_ a given slot (i.e. which byte to start reading from or writing to).
```
mitchmindtree, line 306: > Another thing to consider with tightly-packed structs is that the type `StorageHandler` would then also require an additional field that points to a particular location _inside_ a given slot (i.e. which byte to start reading from or writing to).

I can see how there would need to be different behaviour for `get` and `store` based on whether a struct is packed or  unpacked, however these implementation differences could be resolved at compile-time no?

I guess that would require representing the offset with a zero-sized type (e.g. `StorageHandler<Offset<Bytes3, Foo>>` or similar) which would break a lot of the usability as you could no longer pass the handle to functions that take `StorageHandler<Foo>` :thinking: Maybe that's fine though, as I'd imagine packed types would tend to just store primitive data anyway, so there may be less of a need to pass around `StorageHanlder`s for those fields anyway?

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+Solidity's approach to storage is completely different from the above and has its own problems, as storage variable are hard to identify and storage accesses are not explicit enough. Sway's current approach is also flawed as explained in the [Motivation](#motivation) section.
+
+In Rust, the closest thing to storage accesses is file I/O where a handler is used to read and write to a file similarly to `store` and `get` above.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+- How do we make storage accesses, particularly reads, more ergonomic?
+- Is not having structs tightly packed acceptable? Is that something that we want in the future, potentially via a `#[packed]` attribute that also reorders the fields of the struct for optimal layout?
```
mitchmindtree, line 329: This is a good point - most users will just use whatever the default is so it would be good to get it right.

I guess if we're to pack by default, users could always add their own padding to structs with `[u8; N]` if they want to avoid the awkwardness of typed offsets (mentioned above)?

On the other hand, if we want to track offsets at compile-time, we probably can't add this retroactively without breaking types in existing code, so we'd need to add some field to the handle as you mention. As long as the `StorageHandler` type is opaque, maybe that's fine though?

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+- How do we implement nested dynamic storage types such as `StorageVec<StorageVec<u64>>`?
+- How do we obtain a `StorageHandle` for data stored in a dynamic storage collection such as `StorageMap` or `StorageVec`? Is that even needed/required?
```
mitchmindtree, line 331: This could potentially be quite important, especially for collections of large types?

E.g. Given a `StorageVec<VeryLargeType>`, it would be much cheaper for `StorageVec::get` to return a `StorageHandle<VeryLargeType>` than it would be to return the `VeryLargeType` itself. E.g. given a storage handle, the user could potentially defer the cost of reading the `VeryLargeType` based on some condition in the code.

Perhaps `StorageMap::get` and `StorageVec::get` are better off returning `StorageHandler<T>` rather than `Option<T>`? If we're to go for the suggested `read`/`write` methods for differentiation, usage might look like:
```rust
let elem_handle = storage.my_vec.get(ix); // Get the handle.
```
```rust
let elem_opt = storage.my_vec.get(ix).read(); // Read the `Option<T>`.
```

mitchmindtree (commented):
Nicce, looking much better than the `StorageRef`! I think clearly communicating in our API that these are *not* references to manage expectations of how they behave is a major win, and gives us more time/space to improve on ergonomics and maybe think about the potential of *real* references later on.

rfcs/0008-storage-handler.md
otrho: I'm thinking to prevent sentinel values.  If a value is no longer valid then it should be `None`, not zero, which clearing effectively provides.

rfcs/0008-storage-handler.md
mohammadfawaz: I was actually wondering if we need a separate API to get the handler instead of changing the behavior of `get` since `get` is modelled after the `get` method from Rust's `Vec` and `HashMap`.

Maybe something like this would work:
```rust
let elem_handle = storage.my_vec.get_handle(ix); // or `get_key(ix)`, Get the handle
```
```rust
let elem_opt = storage.my_vec.get(ix); // Read the `Option<T>`
```

rfcs/0008-storage-handler.md
mohammadfawaz: To Toby's point. I see that we do this in the Fuel specs so I followed the same approach here.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
+- Feature Name: Storage Handlers
+- Start Date: 2023-02-10)
+- RFC PR: [FuelLabs/sway-rfcs#0023](https://github.com/FuelLabs/sway-rfcs/pull/023)
+- Sway Issue: [FueLabs/sway#3043](https://github.com/FuelLabs/sway/issues/3043)
+
+# Summary
+
+[summary]: #summary
+
+This RFC introduces the concept of a `StorageHandler` which describes a particular storage slot via its key. The RFC also reworks how storage accesses are represented in the language and how `StorageHandler` can be used to build dynamic storage types, such as `StorageMap` and `StorageVec`, more robustly.
```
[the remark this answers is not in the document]
mohammadfawaz: is `Key` too short? I agree that `StorageHandler` is too long.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+```rust
+pub struct StorageHandler<T> {
+    key: b256,
+}
+```
+
+The type `T` describes the type of the data stored at `key`. Since each storage slot is 64 bytes long, some types cannot fit in a single storage slot. In that case, the data is stored across multiple consecutive storage slots starting at `key`. More on this in [Storage Slot Assignment](#storage-slot-assignment).
+
+Reading and writing the data stored at `key` can be accomplished using the `get` and `store` methods below:
```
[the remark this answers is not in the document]
mohammadfawaz: I'd be happy with `read` and `write`. If anything, to also avoid confusion v.s. `StorageMap::get` and `StorageVec::get`.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+        let key = sha256((key, self.key));
+        get::<V>(key)
+    }
+
+    #[storage(write)]
+    pub fn remove(self: StorageHandler<Self>, key: K) -> bool {
+        let key = sha256((key, self.key));
+        clear::<V>(key)
+    }
+}
```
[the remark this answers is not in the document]
mohammadfawaz: I'm happy to explore that option. The good thing is that the way the user would need to call `insert`, `get`, and `remove` would look the same so if we go one way or another in the future, most end users will not be affected.

rfcs/0008-storage-handler.md
mohammadfawaz: > This is why I'm in favour of not requiring an initialiser either

This is reasonable to me, particularly because writing initializers today, especially for complex structs, is a pain in the behind.

rfcs/0008-storage-handler.md
jtriley-eth: On slot packing, the way solidity handles this is it starts at the first variable and if the current and following variables' sizes are small enough to fit into one slot, it will always be packed. Value types are never split between slots.

```solidity
contract A {
    // slot0[31]
    uint8 a;
    // slot0[32]
    uint8 b;
    // slot1[..]
    uint256 c;
    // slot2[12..]
    address d;
    // slot2[..12]
    uint96 e;
}
```

Since storage reads/writes are generally much more expensive than masking/shifting, in the future it would be good if packed was the default and padded was an option, possibly via a `#[padded]` attribute.

rfcs/0008-storage-handler.md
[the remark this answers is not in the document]
jtriley-eth: Agree, `read` and `write` make more sense here.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+- `s.a` is assigned `sha256("storage_3")`.
+- `s.b.c` is assigned `sha256("storage_3") + 1`.
+- `s.b.d` is assigned `sha256("storage_3") + 2`.
+
+### Empty Types
+
+There are two possible empty types: structs with no fields and enums with no variants. Empty structs are useful for implementing dynamic storage types such as `StorageMap` and `StorageVec`. Storage variables that have empty types should not consume any storage slots.
+
+### Behavior of `StorageKey`
```
jtriley-eth, line 191: Since we're defining `StorageKey` behavior, it seems storing values at different hashes is inefficient. The current behavior creates the need for an additional 32 bytes for each storage variable, which increases all contract sizes. The alternative would be to hash the slot at runtime, adding additional gas to already expensive storage reads/writes.

For comparison, Solidity increments starting at slot zero and increments by one for each variable. Dynamic arrays store their length in the slot and the array starts at `keccak256(slotIndex) + arrayIndex` and hash maps are stored at `keccak256(abi.encodePacked(key, slotIndex))`.

This minimizes bytecode size since the slots start at zero and only use hash digests for dynamic values.

mohammadfawaz: The main reason why packing was not originally considered is because it leads to read-modify-write patterns. Say we have
```
storage {
    x: u64 = 0,
    y: u64 = 0,
    z: u64 = 0,
    w: u64 = 0,
}
```
and we decide to store all the variables in a single storage slots (say at key 0x00..0) since they actually fit there. If I would like to update `x`, then I need to first **read** from `0x00.00`, do some shifting/masking to update only the bytes that `x` occupy, and then **write** back the result. Isn't more expensive than just writing `x` directly when it's in its own slot?

nfurfaro: One thing Im not sure about: Are we charged by the storage slot or by the byte when reading/writing ?
Normally I would assume the former, but I seem to recall some discussion about allowing for reading multiple slots in one read for example.

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+
+# Motivation
+
+[motivation]: #motivation
+
+The current approach for reading and writing storage slots/variables has multiple flaws, despite being quite ergonomic:
+
+1. There is currently no way to reason about the location of a particular storage variable, which makes passing "references" to storage variables impossible. One workaround is manipulating and reading storage slots manually via `std::storage::store` and `std::storage::get`, but this approach is quite unsafe. Alternatively, one could declare a `trait` in a library and implement it in a contract and have its methods access storage variables. Those methods are then used in the library to access storage indirectly. This does not solve the full problem and feels more like a workaround than a real solution.
+1. Dynamic storage types, such as `StorageMap` and `StorageVec`, currently require a "hacky" compiler intrinsic called `__get_storage_key` which is not very well defined and is hard to use. Introducing the concept of a `StorageKey` will helps us completely remove this intrinsic.
```
sdankel, line 19:
```suggestion
1. Dynamic storage types, such as `StorageMap` and `StorageVec`, currently require a "hacky" compiler intrinsic called `__get_storage_key` which is not very well defined and is hard to use. Introducing the concept of a `StorageKey` will help us completely remove this intrinsic.
```

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+> **Note**: 
+> The above is a pseudo-code and is not meant to compile for various reasons.
+
+## Implementation Detail
+
+The typed expression `TyStorageAccess` in the compiler should now return `std::storage::StorageKey` which should store the key chosen by the compiler according to the rules described in [Storage Slots Assignment](#storage-slot-assignment). The rest of the flow is be handled automatically after removing all the unnecessary logic in the compiler for reading and writing storage slots.
+
+## Implementing Dynamic Storage Types using `StorageKey`.
+
+Dynamic storage types such `StorageMap` and `StorageVec` currently use the intrinsic `__get_storage_key` which is not very well defined and has a fragile implementation. With `StorageKey` available and returned directly from a storage access expression (as in `storage.my_map`), we can re-implement methods like `StorageMap::insert`, `StorageMap::get`, and `StorageMap::remove` in a safer and cleaner way as follows:
```
sdankel, line 216: I know very little about `__get_storage_key ` and why it's bad. It would be helpful if there was a section going into a little more detail about why it's important to remove the usage of `__get_storage_key`, since it sounds like a key motivator for this RFC.

I'm wondering:
1. What does it mean to say an intrinsic is not well defined?
1. What is fragile about the implementation?
1. What are the risks or vulnerabilities with using it?
1. Why was it used in the first place?

rfcs/0008-storage-handler.md
```diff
@@ -0,0 +1,338 @@
...
+    #[storage(write)]
+    pub fn write(self, other: T) {
+        std::storage::store(self.key, other);
+    }
+}
+```
+
+Notice that `read` returns an `Option` because it is possible that a particular storage slot is not valid, i.e. has not been written before, in which case `None` is returned. If the type `T` spans multiple storage slots, then the method `read` returns `None` if at least one the storage slots that `T` spans is invalid.
+
+With the `StorageKey` type available, this RFC proposes a redefinition of the meaning of the expression `storage.<var>.<field>...` to return a `StorageKey` "pointing" to the actual data instead of returning the data itself. The RFC also proposes removing the "reassignment" statement `storage.<var>.<field>.. = ..` from the language.
```
sdankel, line 54: What's the reason for removing the reassignment syntax? Why is it problematic to keep it?

If we're keeping the assignment syntax for reading the data, it seems surprising that it won't work for writing. If I can do:

```
let x_key: StorageKey<u64> = storage.s.x
```

I'd expect to be able to do:

```
storage.s.x = 42
```

mitchmindtree: > If I can do:
> 
> let x_key: StorageKey<u64> = storage.s.x
> 
> I'd expect to be able to do:
> 
> storage.s.x = 42

The key difference is that under this proposal, in both of these cases, the types are totally different. E.g. in former, you assign a storage key to a variable of type `StorageKey<u64>`, but in the latter, you attempt to assign an integer to a storage key.

I think the removal of the assignment syntax is also useful for more clearly distinguishing the cost involved in storage interaction vs regular variable/field assignment, e.g. assigning the following two pieces of code have vastly different costs, but it's not at all obvious to a new user reading sway code:
```rust
my_struct.foo.bar = 42;
```
```rust
storage.foo.bar = 42;
```

rfcs/0008-storage-handler.md
mohammadfawaz: I will make the motivation section clearer!

mitchmindtree (approved):
Nice work Mo, LGTM!

It would be nice if we could know we were picking the right default w.r.t. packing fields. I worry that many users will just use whatever the default is - it would be a shame to pick the more costly.

I wonder if we could somehow keep the `b256` value of the key private for now, and clarify in our documentation that the layout of the keys may be subject to change until we decide on the right default?

Alternatively, for now we could require that users decorate struct storage members with `#[unpacked]` so that the current behaviour is explicit? Then in the future, we could remove this requirement once we settle on a default?

nfurfaro:
>I wonder if we could somehow keep the b256 value of the key private for now, and clarify in our documentation that the layout of the keys may be subject to change until we decide on the right default?

Seems reasonable to me. As long as we decide by v 1.0 that should be fine.

otrho (approved):

rfcs/0008-storage-handler.md
mohammadfawaz: Packing is now the default, at least for structs. This offers the most flexibility right now. If you want your variables to be packed, put them in a struct. Otherwise, leave them as their own storage variables.

rfcs/0008-storage-handler.md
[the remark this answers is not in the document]
mohammadfawaz: Went with this option. To your point, it seems that we should allow this in Sway:
```
struct Foo;
impl Vec<Foo> {}
```
even though Rust doesn't.

rfcs/0008-storage-handler.md
mohammadfawaz: Packed is now the default. Future discussion can consider unpacking sturcts via an annotation.

rfcs/0008-storage-handler.md
mohammadfawaz: Packed structs are now the default.

rfcs/0008-storage-handler.md
mohammadfawaz: Went with your suggestion here as well.

rfcs/0008-storage-handler.md
mohammadfawaz: Packing is now the default but only for structs.

rfcs/0008-storage-handler.md
mohammadfawaz: I added a sentence about why its implementation is bad. This is mostly the big reason for removing this. The only way to implement it is bad because it relies on inlining.

rfcs/0008-storage-handler.md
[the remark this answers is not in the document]
mohammadfawaz: `read` and `write` are used everywhere now.

rfcs/0008-storage-handler.md
mohammadfawaz: Keeping initializers for now for reasons related to packing structs.

IGI-111 (approved):

nfurfaro (approved):
Looks great.

esdrubal (approved):

PR #24, opened 2023-02-21
Standardize module structure

IGI-111:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/57018b079d1012e6cc1be3db5cb36db7dd3e6775/rfcs/0009-structured-modules.md)

This RFC proposes changes to standardize the way we define and structure modules in a way more similar to Rust's `mod`.

### DIFF
--- a/rfcs/0006-structured-modules.md
+++ b/rfcs/0006-structured-modules.md
@@ -1,0 +1,190 @@
+- Feature Name: standard_module_structure
+- Start Date: 2023-02-14
+- RFC PR: [FuelLabs/sway-rfcs#24](https://github.com/FuelLabs/sway-rfcs/pull/24)
+- Sway Issue: [FueLabs/sway#4191](https://github.com/FuelLabs/sway/issues/4191)
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes a change to the way defining modules works in Sway.
+
+Instead of requiring programmers to specify a path to their module definitions
+(as in: `dep dir1/dir2/dep_name;`) and a possibly different library name (as in
+`library library_name;`), this proposes to have the programmers specify a
+single name for a module through the Rust-like syntax of `mod module_name;` and
+have the path to the file containing the associated source code and the library
+name used for imports be implicitly defined.
+
+# Motivation
+
+[motivation]: #motivation
+
+The current `dep` based import system is confusing and possibly error prone in
+that it allows the definition of surprising and unpredictable module structures
+that can be completely detached in name from how those modules are imported.
+
+This permissive scheme also allows namespace collisions wherein two separate
+packages or modules can be imported despite having the same name.
+
+The proposed change should formalize a predictable module structure, reduce the
+number of editions required when refactoring a module and eliminate the
+possibility of namespace collisions between modules of the same package.
+
+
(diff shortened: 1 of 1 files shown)


nfurfaro:
While this RFC doesn't directly address adding the ability to re-export modules (i.e: the `pub use my_mod;` syntax), it seems like it might set us up better to do so in the future... if so then I'm all for it.

rfcs/0009-structured-modules.md
```diff
@@ -0,0 +1,190 @@
...
+We also do not consider the visibility of modules here and reproduce the
+public by default scheme of `dep`.
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+We could consider eliminating module type headers like `library;`, `contract;`
+and `predicate;` from the source altogether and defining the exports in the
+`Forc.toml` or implicitly through usage.
```
mitchmindtree, line 184: Huge fan of eliminating these from src files in favour of assuming `library` by default, and declaring the top-level program type in the forc manifest. Currently forc basically has to do a parse of the file just to check the program type (it's not safe to just grab the first line as a user might have combinations of whitespace, comments and now module-level comments preceding the declaration). But yes, makes sense for this to be addressed as follow-up.

rfcs/0009-structured-modules.md
```diff
@@ -0,0 +1,190 @@
...
+
+The elimination of file type headers such as `library;` or `contract;` when
+they aren't needed is not considered here. Neither are wider issues about the
+structure and type of package exports. But clarifying the module structure will
+definitely give us a more stable base to consider these.
+
+We also do not consider the visibility of modules here and reproduce the
+public by default scheme of `dep`.
+
+# Future possibilities
```
mitchmindtree, line 178: Another future possibility to consider is that we'll likely want some qualifier for referring to the root module of the current package, similar to rust's `crate` keyword. There was some mention of this in #8 too [here](https://github.com/FuelLabs/sway-rfcs/pull/8/files#diff-bb987111424978b661aa08f82a8260cbfe96ecc1eff267876dd623f33a8430b1R294-R301).

mitchmindtree (approved):
Love it - nice and simple and to the point. I think this makes for a nice start on improving our current module sitch!

rfcs/0009-structured-modules.md
```diff
@@ -0,0 +1,190 @@
...
+
+[unresolved-questions]: #unresolved-questions
+
+The elimination of file type headers such as `library;` or `contract;` when
+they aren't needed is not considered here. Neither are wider issues about the
+structure and type of package exports. But clarifying the module structure will
+definitely give us a more stable base to consider these.
+
+We also do not consider the visibility of modules here and reproduce the
+public by default scheme of `dep`.
```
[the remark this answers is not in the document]
IGI-111: Yeah this should probably be the very next order of business for modules after this.

mohammadfawaz (approved):
Excellent work. I say we do this sooner than later since it's a breaking change.

otrho (approved):

JoshuaBatty (approved):
This is much nicer than our current approach. Keen to see this land.

PR #25, opened 2023-04-05
Private modules and reexports

IGI-111:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/21b3ba2b137e1188fb066bdf8ec5e564d8d6f1bb/rfcs/0008-private-modules.md)

This RFC proposes to make symbols private by default and to introduce reexports.

### DIFF
--- a/rfcs/0008-private-modules.md
+++ b/rfcs/0008-private-modules.md
@@ -1,0 +1,212 @@
+- Feature Name: private_modules
+- Start Date: 2023-04-05
+- RFC PR: [FuelLabs/sway-rfcs#0007](https://github.com/FuelLabs/sway-rfcs/pull/25)
+- Sway Issue: [FueLabs/sway#4446](https://github.com/FuelLabs/sway/issues/4446)
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes introducing two changes.
+
+First, instead of exposing the entire module structure by default and
+modulating the privacy of elements globally and one by one, this RFC proposes
+to make all modules private by default and only allow importing symbols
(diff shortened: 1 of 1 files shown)


rfcs/0007-private-modules.md
```diff
@@ -0,0 +1,212 @@
...
+
+* if an item is public, then it can be accessed outside of a module `m` if you
+can access all the item's ancestor modules from `m`. You may also be able to
+name the item through reexports.
+* if an item is private, then it can be accessed by the current module and
+its descendants.
+
+## Reexports
+
+We allow publicly reexporting items using the `pub use` syntax. This allo the
```
otrho, line 62:
```suggestion
We allow publicly reexporting items using the `pub use` syntax. This allows the
```

otrho (approved):
We may initially want to put this behind an 'experimental' flag and issue a warning if it's off, and set a date to then make it the default.  For smaller projects it's pretty easy to add a `pub` where required, but bigger projects might not be happy about being forced to do so with no warning.

But otherwise :+1:

mohammadfawaz (approved):

JoshuaBatty (approved):

kayagokalp (approved):
LGTM. 

I am leaving this as a note to the implementor or to future me to fix it:

Not sure if this is going to be effected by this implementation, we are injecting contract ids into custom namespaces named after the relevant contract during `forc build`. If modules are going to be private by default, we might need to make sure that those injected contract ids are still visible from root namespace. IIRC we should have some tests in place that should prevent that regression but better be safe than sorry!

mitchmindtree (approved):

PR #26, opened 2023-04-06
External Transaction representation in scripts

SwayStar123:
This RFC proposes a way to represent external third part transactions inside scripts to allow safe and trustless MEV extraction that can benefit both the user and the searcher.

### DIFF
--- a/rfcs/0007-multisigner-scripts.md
+++ b/rfcs/0007-multisigner-scripts.md
@@ -1,0 +1,128 @@
+- Feature Name: Multi-signer Scripts
+- Start Date: 2023-04-07
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/001)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+
+# Summary
+
+[summary]: #summary
+
+This RFC proposes a new "Tx" struct/enum to be used in scripts to allow for atomic execution of bundle-like scripts.
+
+Allowing programmers to insert external transactions into their own scripts for atomic execution without any additional security assumptions can allow for a much more decentralised and secure MEV environment where both users and searchers benefit greatly, whilst also unlocking previously unseen types of MEV.
+
+# Motivation
+
+[motivation]: #motivation
+
+The current script features allow programmers to execute multiple transactions atomically, but it is limited to one party and cannot be extended to allow for atomic execution of transactions from multiple parties. Allowing insertion of third party transactions into scripts would remove the need for centralised block proposers who have privileged access to signed orderflow.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The "Tx" enum comes in two variants, "Signed" and "Unsigned". The Signed variant represents a Transaction which has already been signed, and the Unsigned variant represents an intent of a transaction, without actually being signed yet.
+
+A script including a Signed Tx may execute the signed tx once at any point in their script, creating their own contract calls before or after the external tx. This can be used to arbitrage price changes, provide just in time liquidity, or other sorts of MEV tactics.
+
+A script including an Unsigned Tx may also execute the unsigned tx at any point in their script, however, before submitting the script to the blockchain, it must be sent to the third party so they can sign and fulfill their intent, or else the script will revert.
+
+Eg.
+
+```sway
+script;
+
+fn main(third_party_tx: Tx) {
+    // Arbitrary actions before execution of third party tx
+    ...
+
+    // Execute the third party tx
+    third_party_tx.execute()
+
+    // Arbitrary actions after execution of third party tx
+    ...
+}
+```
+
+You could also inspect the third party tx in order to make decisions based on the third party tx
+
+```sway
+script;
+
+use exchange_abi::Exchange;
+use std::constants::ZERO_B256;
+
+fn main(third_party_tx: Tx) {
+    let exchange_contract_address = 0x9299da6c73e6dc03eeabcce242bb347de3f5f56cd1c70926d76526d7ed199b8b;
+    let caller = abi(Exchange, exchange_contract_address);
+
+    // Retrieve calls made to the given caller by the external tx
+    let calls: Vec<Calls> = third_party_tx.inspect(caller.swap);
+
+    // Amount swapped by third party tx in the last call to swap (assuming there is a call to swap)
+    let swap_call = calls.pop().unwrap();
+    let amount = swap_call.coins;
+    let asset_a = swap_call.asset_id;
+    let asset_b = swap_call.arguments.asset_b;
+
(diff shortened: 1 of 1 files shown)


rfcs/0007-multisigner-scripts.md
```diff
@@ -0,0 +1,128 @@
...
+
+The current script features allow programmers to execute multiple transactions atomically, but it is limited to one party and cannot be extended to allow for atomic execution of transactions from multiple parties. Allowing insertion of third party transactions into scripts would remove the need for centralised block proposers who have privileged access to signed orderflow.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The "Tx" enum comes in two variants, "Signed" and "Unsigned". The Signed variant represents a Transaction which has already been signed, and the Unsigned variant represents an intent of a transaction, without actually being signed yet.
+
+A script including a Signed Tx may execute the signed tx once at any point in their script, creating their own contract calls before or after the external tx. This can be used to arbitrage price changes, provide just in time liquidity, or other sorts of MEV tactics.
```
matt-user, line 26:
```suggestion
A script including a Signed Tx may execute the signed tx once at any point in their script, creating their own contract calls before and/or after the external tx. This can be used to arbitrage price changes, provide just in time liquidity, or other sorts of MEV tactics.
```

rfcs/0007-multisigner-scripts.md
```diff
@@ -0,0 +1,128 @@
...
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+The current alternatives are 
+
+-Regular scripts. While scripts already allow atomic execution of multiple contract calls, they are not an effective alternative for bundles as external transactions have no representation in scripts.
+-Bundles. Transactions inside bundles can easily be unbundled and exploited by privileged actors such as validators.
+
+Without representation of external transactions inside scripts, searchers would have to have to send bundles of scripts to validators for execution, where validators can easily ignore some scripts and replace them with their own, exploiting the searcher. This sort of attack has happened multiple times in the past.
+
+Representing external transactions in scripts allows a searcher to sign an entire script at once, where tampering with the contents of the script would render it invalid, safegaurding the searcher from unbundling attacks or uncle bandit attacks.
```
matt-user, line 105: Maybe I missed it, but what would be the actual mechanism for invalidating the script if this were to happen.

SwayStar123: Thats just how scripts normally work, a keypair signs an entire script, if the bytecode changes then the signature wont match. Similarly for inclusion of Unsigned Txs, if the Unsigned Tx never gets signed and the programmer tries to submit it to the blockchain, it would be an invalid script as there would be a missing signature

Voxelot:
This should be more descriptive in terms of the changes to the transaction format in fuel-specs that would be required: https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/tx_format/index.md

For example, not all txs fit squarely into signed vs unsigned if they use predicates instead.

If the sole-purpose of this is for MEV bundling, maybe it would be simpler to add some APIs to fuel-core that allow MEV searchers to propose entire blocks? Otherwise you'll be more likely to run into the per-tx limits such as

- MAX_GAS_PER_TX
- MAX_INPUTS
- MAX_OUTPUTS
- MAX_WITNESSES
- MAX_SCRIPT_LENGTH
- MAX_SCRIPT_DATA_LENGTH

(see https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/tx_format/constants.md)

This would also have a negative impact on fuel's parallelism, as txs that could be run in parallel would now be run as a sequentially determined batch.

It's also not clear how inputs would be authorized for spending on unsigned scripts. For example, if I submit a dex trade such as DAI->USDC using coin inputs that I own, how would I prevent the MEV searcher from stealing the funds or using them in a totally different way?

SwayStar123:
> This should be more descriptive in terms of the changes to the transaction format in fuel-specs that would be required: https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/tx_format/index.md
> 
> For example, not all txs fit squarely into signed vs unsigned if they use predicates instead.

I dont fully understand how predicates work especially behind the scenes, but if I understand correctly, predicates have no side effects so representing them inside of a script for this usecase would not really make any sense, and since any spending of UTXOs occur inside of a script, if the third party tx (script) spends any predicate UTXOs it should not really matter here

> 
> If the sole-purpose of this is for MEV bundling, maybe it would be simpler to add some APIs to fuel-core that allow MEV searchers to propose entire blocks? Otherwise you'll be more likely to run into the per-tx limits such as
> 
> * MAX_GAS_PER_TX
> * MAX_INPUTS
> * MAX_OUTPUTS
> * MAX_WITNESSES
> * MAX_SCRIPT_LENGTH
> * MAX_SCRIPT_DATA_LENGTH
> 
> (see https://github.com/FuelLabs/fuel-specs/blob/master/src/protocol/tx_format/constants.md)

I dont know if this is technically feasible but is it not possible for the VM to treat the external txs represented in the script as a seperate tx? The problem with allowing searchers to propose blocks is that it gives them all the power, without any power being held in the hands of the user the searchers can exploit any mempool transactions to the fullest with no benefit to the user but only to the searcher and the sequencer. There needs to be some way for the user to withhold execution or else they wont see a penny of the value of their orderflow. Which is why i suggest an intent + multi sig based approach. Users can still just submit signed txs to the mempool if they dont care about any of this.

> 
> Bundling into a single tx would also have a negative impact on fuel's parallelism, as txs that could be run in parallel would now be run as a sequentially determined batch.

Any kind of bundle would require sequential transaction processing anyway. Its not possible to frontrun or backrun a tx without its effect occurring, unless I misunderstand your point.
> 
> It's also not clear how inputs would be authorized for spending on unsigned scripts. For example, if I submit a dex trade such as DAI->USDC using coin inputs that I own, how would I prevent the MEV searcher from stealing the funds or using them in a totally different way?

Not sure I understand your question. If your intent (unsigned tx given to searchers) is never signed and is attempted to be executed the client should automatically just reject such a transaction as being invalid, as an unsigned transaction cannot be executed.

If you submit a signed tx to the mempool, any searcher can exploit your transaction by frontrunning or backrunning it (assuming theres some sort of agreement between them and the block producer, like on ethereum with flashbots). Thats the status quo.

What im proposing is that users may submit unsigned transactions (Intents) to searchers, who can build scripts with them to extract whatever MEV they can find. Whichever script has the highest bid can be signed by the user to fulfill their intent, which can then be submitted to the mempool. As the MEV will theoretically be extracted by this point with a fee paid to the user for the privilege (instead of to a miner or block producer), there wont be anything left to exploit. Ofcourse due to inefficiencies there might still be some sort of MEV left that can be extracted but I dont think thats worth worrying about.

Hope that clears up that question, if you think my explanation of the above in the RFC was inadequate let me know

SwayStar123:
Closed in favour of https://github.com/FuelLabs/rfcs/pull/16

PR #30, opened 2023-08-07
Expressive diagnostics

ironcev:
This RFC calls for improvements in Sway warning and error messages.


rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
+- Feature Name: `expressive_diagnostics`
+- Start Date: 2023-08-07
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/001)
```
kayagokalp, line 3:
```suggestion
- RFC PR: [FuelLabs/sway-rfcs#0010](https://github.com/FuelLabs/sway-rfcs/pull/030)
```

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
```
kayagokalp, line 4: Is there a sway issue for this RFC?

kayagokalp (changes requested):
Before diving into the content, leaving these as a note so that we do not skip them later on. 

Also i think file name should be `0010-expressive-diagnostics.md` :)

rfcs/0000-expressive-diagnostics.md
ironcev: As far as I understood the RFC process, the name gets renamed and the links get set after the PR is accepted. The PR accepts commits after it is approved so I'll adjust the links and the file name once we agree on the content. So, dive in, looking forward to your comments :-)

ironcev: I'll also create the Sway issue before adjusting the RFC.

rfcs/0000-expressive-diagnostics.md
ironcev: Thanks for the suggested change! I hope this is exactly how it will look like at the end :-) but let's stick to the process and see if the RFC will get accepted.

rfcs/0000-expressive-diagnostics.md
kayagokalp: Ahh okay, my bad then!

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+
+[unresolved-questions]: #unresolved-questions
+
+I expect to get feedback on the Diagnostic API, most importantly, if the suggested elements, _Code_, _Reason_, _Issue_, _Hints_, and _Help_ cover all of the cases we've encountered so far.
+
+One open question is the usage of symbolic codes like e.g. `non_camel_case_types` instead of, as proposed, e.g., `W0001`.
+This would allow us more readable `#[allow]` attributes. E.g.:
+
+```Rust
+#[allow(non_camel_case_types)]
```
sdankel, line 289: This is much better!

ironcev:
One thing to agree on will be the guidelines for structuring messages, wording etc.. Here are three possible approaches (named by me, just that we can distinguish them):

## The Elm approach

The one explained in the [Compiler Errors for Humans](https://elm-lang.org/news/compiler-errors-for-humans) article. Rich, friendly English, full sentences. Breaking the issue down in details through a story.

image

This is the approach I opted for so far in errors and warnings developed with the new `Diagnostic`. It is also the one suggested in the [Concepts Error Messages for Humans](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2429r0.pdf) paper.

Pros:
- friendly, detailed, allows enough of context

Cons:
- might be too wordy for experienced programmers (mitigated by the fact that the first, short, line should be enough ;-))

## The Rust approach

Short, sometimes very technical, stripped of official grammar, focused on the point. E.g.:

image

Pros:
- right to the point, communicating only the bare minimum, no clutter in wording, perfect for experienced programmers

Cons:
- might be still too difficult to understand for non-experienced programmers

## Fluent Rust approach

As suggested by @xunilrj in [this comment](https://github.com/FuelLabs/sway/pull/4965#issuecomment-1682193441). The idea is  that text in code snippet that is in the same color makes a readable sentence.

image

Pros:
- the best of two approaches above

Cons:
- might be difficult to always get the top-down reading; depending on the error, the pieces of code might be arbitrary scattered away and difficult to figure out how to order.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+[guide-level-explanation]: #guide-level-explanation
+
+When using `forc` via command line Sway programmers will get detailed diagnostics consisting of the following elements:
+
+- _Level_: Error or Warning.
+- _Code_: Unique code of the diagnostic.
+- _Reason_: Short description of the diagnostic, not related to a specific issue that the programmer did. The _reason_ answers the question "Why is this an error or warning?" E.g., Because - "Constants cannot be shadowed". It points out the general language rule that was violated.
+- _Issue_: Short description of the concrete issue caused by the programmer. E.g., "Variable "X" shadows imported constant with the same name"
+- _Hints_: Detailed descriptions of the diagnostic. Hints are placed in the source code. They point to the problematic line of code but also to other places in code that can help to understand the issue.
+- _Help_: Additional friendly information, which is not related to a place in code, that helps understanding and solving the issue.
```
IGI-111, line 72: Which of these are optional, which of these are mandatory?

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+
+For _hints_, not being in source code means that they are not rendered at all. Diagnostics API will ignore them and clients (CLI, LSP) will not be aware of them.
+
+For the _issue_, clients will always display it as a part of diagnostic description. If it is in code, clients can choose to additionally display it in code.
+E.g., `forc` always shows the _issue_ text as a part of the diagnostic description, but if the _issue_ is in code and there is a _hint_ pointing to the same place in code, the _issue_ will not be rendered as a _label_ in code. This overloading of the _issue_ text by a _hint_ is visible on the above example and is enforced via Diagnostics API. It gives us the flexibility to phrase the diagnostic description and the hints in an independent way, while allowing backward compatibility with the current warnings and errors that have only a single message and span.
+
+Using `Span::dummy()` to denote non-existance of an element is in line with the current usage of `Span::dummy()`. It also allows declarative definition of all the _hints_, knowing that some of them might not be shown, depending on the concrete `Span` value passed to the enum variant.
```
IGI-111, line 210: I know that the LSP team is currently wondering if we could guarantee all spans have proper paths, so encouraging use of dummy spans may not be a good idea. We may want to use Option here.
@JoshuaBatty what do you think?

IGI-111 (approved):
A few nits that require clarification, but LGTM

rfcs/0000-expressive-diagnostics.md
ironcev: At the moment, for backward compatibility, only the _Level_ and the _Issue_. And since switching to `Diagnostic` will take time, it will stay like that for a while.

Once all errors and warnings support `Diagnostic` IMO these should be mandatory:
- Level
- Code
- Reason
- Issue

rfcs/0000-expressive-diagnostics.md
ironcev: TL:DR: I am all for `Option` but that would IMO mean getting rid of `Span::dummy` in general, or have a cumbersome checks and code every time we `emit_err` or `emit_warn`. 

In my first commit I left the comment saying something along the lines _// Should we use Option here?_. Strictly speaking, it should be an `Option`.

But after looking in the existing usages, it is clear that `Span::dummy()` is used to communicate non-existance. If we decide to use `Option` here, we should consistently use it everywhere. Which opens the question of the role of `Span::dummy()`.

Also, there is a good deal of convenience in using `Span::dummy()` in diagnostics because the existence of `Span` is the regular case, overwhelmingly regular. With `Span::dummy()`s we don't need to write `Some(my_span)` every time we define warning/error in code, but just `my_span`. Same with unwrappings etc. In those rare cases we know that we don't have span `Span::dummy()` sticks out nicely, better then `None` ;-)

Last but not  least, `Span::dummy()`s get propagated through the code now. When they reach diagnostic, we can just pass them over. If we don't switch everywhere to `Option<Span>` that would mean always checking if we got `Span::dummy()` and turning it into `None`. E.g. `Span::join_all()` returns `Span;:dummy()` if there are no spans.

sdankel:
(x-posting from slack) In general I think we should trending towards fewer types of compiler warnings with fields for context/customized hints, rather than having many compiler warnings that basically mean the same thing but with a slight difference in context.
The large number of `DeadX` warnings for dead code is an example of this. In my perfect world, there would just be 1 DeadCode warning that has generic context fields for the type of dead code and other info, like the ident and verb.

I'm not sure if this is called out, but we should have a hosted documentation of all error codes, similar to rust's [error code index](https://doc.rust-lang.org/error_codes/error-index.html). Every error diagnostic should link to the error code documentation. In my view, we should keep the compiler diagnostic output as short and simple as possible, with the detailed explanation of the error in the hosted docs. We shouldn't need an extra field for this since it would only rely on the mandatory error code.

If we're aligned on this, there's another clear benefit to having fewer error codes: less documentation to maintain!

rfcs/0000-expressive-diagnostics.md
sdankel: The nice thing about Option<Span> is that we could be explicit about which Warnings & Error must have a real span. Without this, bugs can easily creep in that are only surfaced in LSP. If we don't have tests for that particular edge case, they can be glaringly obvious to the end user.

My understanding is that most Warnings and Errors _should_ have a span so it would only need to be optional in some rare cases, but I could be wrong about that. It might be worth doing an audit of all the existing diagnostic types to see how many should _not_ require a real span.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+Programmer will have the option to open the full compiler diagnostic, getting the same output like when using `forc` from the command line, similar to this Rust example:
+
+![oopsies image not showing](../files/0010-expressive-diagnostics-vscode-full-diagnostic.png)
+
+## Wording guidelines
+
+To ensure consistency and apply best practices<sup>[5]</sup> the diagnostics will follow these guidelines:
+
+- _Reason_ will be short and point out the language constraint which was violated. It will _not_ finish in fullstop, to emphasize brevity.
+- _Issue_ will be short and point out the specific situation encountered in code. It will also _not_ finish in fullstop, to emphasize brevity.
```
sdankel, line 91: What is the difference between these two? It feels a little redundant to have both.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+- CLI or IDE plugin might choose to combine the _reason_ and the _issue_ as shown at the image above.
+- _Reason_ and _issue_ are given in plain english language, free of, e.g., compiler jargon.
+- _Hints_ and _help_ are as well written in plain english, using proper punctuation and grammar rules.
+- _Hints_ and _help_ try to give as much of useful context as possible and to be as specific as possible.
```
sdankel, line 95: What would go in `help` that could't be expressed in `hints`? For example, in the `cannot be shadowed` example, it says that constants cannot be shadowed multiple times in slightly different ways. The longer the message, the harder it is to digest, so we should just say it once at the top, and have a hint about how to fix it.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+When using `forc` via command line Sway programmers will get detailed diagnostics consisting of the following elements:
+
+- _Level_: Error or Warning.
+- _Code_: Unique code of the diagnostic.
+- _Reason_: Short description of the diagnostic, not related to a specific issue that the programmer did. The _reason_ answers the question "Why is this an error or warning?" E.g., Because - "Constants cannot be shadowed". It points out the general language rule that was violated.
+- _Issue_: Short description of the concrete issue caused by the programmer. E.g., "Variable "X" shadows imported constant with the same name"
```
sdankel, line 70: Imo this would be better as a hint, pointing to where variable X is defined, since it's directly referencing the code. I think anything that is referencing specific code should be placed pointing to the code it references.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+Hints:  [error] This is the enum "MyName" that has the same name as the imported struct.
+        [info]  The struct "MyName" gets imported here.
+        [info]  This is the original declaration of the imported struct "MyName".
+Help:   Items like structs, enums, traits, and ABIs must have a unique name in scope.
+        Consider renaming the enum "MyName" or using an alias for the imported struct.
+```
+
+## Diagnostic codes
+
+A diagnostic _code_ uniquely identifiy a particular _reason_. Considering that we already have these diagnostic areas:
```
sdankel, line 119:
```suggestion
A diagnostic _code_ uniquely identifes a particular _reason_. Considering that we already have these diagnostic areas:
```

sdankel, line 119: Can the same reason be used as both an error and a warning? Like in some contexts it's only a warning, but in other contexts it's actually a blocker to compilation? 

For instance, if the compiler ever had a "strict" mode, where some warnings would be treated as errors.

rfcs/0000-expressive-diagnostics.md
```diff
@@ -0,0 +1,356 @@
...
+- increased efficiency in resolving compiler errors and warnings.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+When using `forc` via command line Sway programmers will get detailed diagnostics consisting of the following elements:
+
+- _Level_: Error or Warning.
+- _Code_: Unique code of the diagnostic.
```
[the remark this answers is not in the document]
sdankel: Never mind, I see that's explained below.

rfcs/0000-expressive-diagnostics.md
JoshuaBatty: Yeah agree with Sophie. Have an Option instead of a Dummy span would be a lot better for the language server. We were running into a bug a few weeks ago where no diagnostics at all where showing in the editor. Turned out this was due to one of the tokens was assigned a DummySpan in the compiler. This means that the Span range was 0,0,0,0. If an LSP client receives a vector of Diagnostics and just one of them has a range of 0,0,0,0 then the client throws the entire Vec away.

ironcev:
> I think we should trending towards fewer types of compiler warnings with fields for context/customized hints

Agree completely. Expressive means exactly that. One (preferably) warning/error with specific contextual explanation.

As with everything, sometimes adding a new error code will be easier then trying to customize hints, but as a general rule, deffinitely yes: less error code, more contextual explanation.

> I'm not sure if this is called out, but we should have a hosted documentation of all error codes, similar to rust's error code index.

Yes, the RFC calls for this as well. That's one of the benefits comming from diagnostic codes, as pointed in the RFC, possibility to link it to the documentation or/and to have `forc explain --ABC` command.

> In my view, we should keep the compiler diagnostic output as short and simple as possible

Here we need to strike a balance between the three options given [above](#issuecomment-1682280094). I do see the examples I did so far, Elm inspired as too wordy. But Rust as too scarce. I see Sway targetting larger programmer's audience, many of whome will need more guidance.

My proposal here is to collect some more different examples and then see how they could look like in different approaches. Since I anyway have few issues in this area, that cover different things, I propose I demonstrate alternatives once the errors/warnings are implemented.

rfcs/0000-expressive-diagnostics.md
ironcev: I was playing with existing diagnostics a bit and think you are right. We should try to communicate primarily through code snippets and _hints_.

There might still be cases where message at the top will be more suitable (at the moment for backward compatibility we support messages without spans) but this should be an exception and not a rule.  

Putting a TODO below to provide examples so that we can compare approaches.

rfcs/0000-expressive-diagnostics.md
ironcev: _identifes_ is surely a step forward compared to _identifiy_ but still not reaching the spelling we want to have :smile: 

I'll spell check the whole document and provide a commit.

ironcev: Great point! Treat warnings as errors is my preferred way of compiling :-)

The easiest way to get that functionality now, would be to solve it at the final receiver of the diagnostics, before the next step, whatever that is. Mean, compiler emits diagnostics and before we generate IR we treat warnings as errors.

`#[allow()]`s would be handled in the compiler before, at the emitting time, of course.

Other way around, treating errors as warnings, I guess we don't have a use case for that.

That means, yes, we can treat errors as warnings without changing their definitions, should be straightforward.

ironcev:
Thanks for the feedback so far!

Below is my homework out of your feedback :-)

The first point is crucial. It will be easier to discuss approaches if we can compare them. 

- [ ] Provide alternative examples for representative error messages:
  - [ ] Remove _issue_ from the top and place it in code.
  - [ ] Reduce amount of information in _help_, formulate it in _hints_.
  - [ ] Demonstrate how the _Fluent Rust approach_ might look like.
- [ ] Propose how to map error code like `W0001` to human readable like `non_camel_case_types` which can be used in `#[allow()]`s once we have them. E.g., `#[allow(non_camel_case_types)]`
- [ ] Document that warnings can be treated as errors.
- [ ] Extend guidelines:
  - [ ] Introduce new error codes restrictively. Reuse existing error codes. Use _hints_ and _help_ to communicate specific cases.
  - [ ] Use _help_ rarely. Prefere _hints_. Leave _help_ for longer suggestions.

sdankel:
I appreciate your thoughtful responses and the compiled list of action items!

Before implementing the new Diagnostic types, it would be ideal to document of all of the existing Error/Warning types mapped to the new Diagnostic type. It will be faster & easier to provide feedback on the documentation rather than individual PRs. To your point, it's easier to draw conclusions when looking at concrete examples.

It looks like we have 30 Warnings and 163 errors. I suspect we could cut down the number of errors significantly. Here are some classes of errors that could probably be represented with 1 diagnostic type.

- `DeclIsNotAX` and `NotAX`
- `ImportPrivateX`
- `UnknownX`
- `MultipleDefinitionsOfX`
- `Unimplemented` and `UnimplementedWithHelp` - should there always be a help?
- `XNotAPartOfInterfaceSurface`
- `DoesNotTakeTypeArgumentsX` and `TypeArgumentsNotAllowed`
- `ImmediateXTooLarge`
- `UnknownType` and `UnknownTypeName`
- `RecursiveType` and `RecursiveTypeChain`
- `XFromExternalContext`

As soon as we add error codes, it will become harder to change/remove error types, so I think it's very important to review these up front.

ironcev:
Well, I appreciate your detailed feedback :bowing_man: :smile: It helps a lot. Thumb up for the upfront review, here is the issue to collect and track the proposals: https://github.com/FuelLabs/sway/issues/4985

ironcev:
@sdankel @JoshuaBatty @xunilrj All the bullet points listed [above](#issuecomment-1684135754) are integrated into the RFC. Thanks once again for your feedback!

The comparison of the two approaches for formulating error messages can be seen [here](https://github.com/FuelLabs/sway/issues/5056#issuecomment-1698964837).

After trying them I am definitively for the formulation proposed by Sophie and I integrated it into the RFC:
- only _reason_ in the header
- _issue_ in code
- _help_ preferably in code if possible
- simpler sentences, right to the point

Regarding the fluent syntax, despite its appeal, it doesn't fit to diagnostics because of the following reasons:
- It's difficult to get a fluent message in a general case. E.g., a shadowed constant can be both above and below the variable that shadows it. This completely changes the message. In an arbitrary case, getting the order right might be even impossible, even if we could structure a fluent message.
- Having different parts in code in the color of the warning or error, e.g., yellow or red, is actually confusing because it blurs the actual line which causes the error. It is not immediately clear which part of the fluent sentence actually causes the warning/error and which is the context.
- It doesn't play well with other clients of the diagnostics. E.g., in VS Code hint and VS Code Problems only disconnected fragments of the sentence would be visible, without the code they relate to. This makes them hard do understand and actually unusable.

Can you please accept the RFC or provide additional remarks if you have some?

sdankel (approved):
This looks great! I hadn't thought about the VSCode problems view, but that is certainly a good thing to pay attention to. Thanks for your hard work on this!

xunilrj (approved):

JoshuaBatty (approved):
Great work on this.

esdrubal (approved):

anton-trunov:
@ironcev Sorry, I'm a bit too late to the table, however, do you have any plans to have different levels of diagnostics verbosity that Sway reports to the user?

For instance, a `Forc.toml` option could govern this and less experienced developers could have it at the, let's say, `diagnostics_verbosity = standard` and more experienced devs would set it to something like `diagnostics_verbosity = terse`.

This could probably be useful for CLI error reporting.

ironcev:
>  I'm a bit too late to the table

It's never too late to join the discussion 😄 Thanks for your feedback @anton-trunov!

> do you have any plans to have different levels of diagnostics verbosity that Sway reports to the user

@sdankel pointed in that direction in our Slack discussion, but in a much more advanced way - having diagnostics that can be replaced by a single one depending on the verbosity level etc. We concluded that it would very likely not pay off.

Hiding some parts of the diagnostics is on the other hand trivial and we can easily implement it.

I say let's roll the diagnostics out and get the feedback. If experienced devs would like to have a terse output we can always add it it will be trivial.

My gut feeling, experienced devs will automatically look only for the yellow/red lines and visually ignore everything else. And if those lines do not immediately point to the solution, they will anyhow be happy to have the hints around.

And for the beginners, having expressive diagnostics with rich explanations will be one of the efficient ways to actually learn the language or at least part of it ;-)

anton-trunov:
Thanks @ironcev. Sounds like a great plan!

PR #31, opened 2023-08-17
References RFC

### LINKED ISSUE #28: Write an RFC to clarify the status of references and pointers in Sway
Memory ownership, references and pointers have some pitfalls, not least `ref mut` syntax. We ought to clarify how reference values work in Sway, and perhaps even do away with pointer types altogether.

IGI-111:
Closes #28 

[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/369c7c4e22ac5c889f2e925e52f4258c4c3d0040/rfcs/0010-references.md)

### DIFF
--- a/rfcs/0010-references.md
+++ b/rfcs/0010-references.md
@@ -81,12 +81,12 @@ Pointers can be obtained by using the `__addr_of` intrinsic and dereferenced usi
 ```sway
 let val = 1u64;
 let ptr: *mut u64 = __addr_of(val);
-let ptr_val = __deref(val);
+let ptr_val = __deref(ptr);
 assert_eq(val, ptr_val);
 
 let val = Box::new(1u64);
 let ptr: *mut Box<u64> = __addr_of(val);
-let ptr_val = __deref(val);
+let ptr_val = __deref(ptr);
 assert_eq(val, ptr_val);
 ```
 
--- a/rfcs/0010-references.md
+++ b/rfcs/0010-references.md
@@ -76,19 +76,26 @@ and produces the desired memory representation.
 There are cases where one may want to directly manipulate memory addresses with pointer arithmetic,
 to allow for this we have a pointer type that represents a single address: `*mut T` where T is the type being pointed to.
 
-Pointers can be obtained by using the `__addr_of` intrinsic and dereferenced using the `__deref` intrinsic, like so:
+Pointers can be obtained by using the `__addr_of` intrinsic on a reference and dereferenced using the `__deref` intrinsic, like so:
 
 ```sway
-let val = 1u64;
-let ptr: *mut u64 = __addr_of(val);
-let ptr_val = __deref(ptr);
+let val: u64 = 1;
+let ptr: *mut u64 = __addr_of(Box::new(val));
+let ptr_val: u64 = __deref(val);
 assert_eq(val, ptr_val);
 
-let val = Box::new(1u64);
-let ptr: *mut Box<u64> = __addr_of(val);
-let ptr_val = __deref(ptr);
+let val: Box<u64> = Box::new(1);
+let ptr: *mut Box<u64> = __addr_of(Box::new(val));
+let ptr_val: Box<u64> = __deref(val);
 assert_eq(val, ptr_val);
+
+let val: Box<u64> = Box::new(1);
+let ptr: *mut u64 = __addr_of(val);
+let ptr_val: u64 = __deref(val);
+assert_eq(val, Box::new(ptr_val));
 ```
+
+Dereferencing an invalid pointer is Undefined Behavior.
 
 ## Slices
 
@@ -309,6 +316,9 @@ check to make sure that `const` is only used with reference types.
 ## Pointers
 
 `*mut` should provide a fully typed alternative to `raw_ptr`.
+
+`__addr_of` will have to be altered to return a `*mut` and we'll need to
+introduce `__deref`.
 
 We will not check pointer mutability at this time, hence the `mut` in `*mut`,
 however we'll probably want to introduce `*const` once typed pointers are
--- a/rfcs/0010-references.md
+++ b/rfcs/0010-references.md
@@ -1,7 +1,7 @@
 - Feature Name: `references`
 - Start Date: 2023-08-17
 - RFC PR: [FuelLabs/sway-rfcs#28](https://github.com/FuelLabs/sway-rfcs/issues/28)
-- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+- Sway Issue: [FueLabs/sway#5063](https://github.com/FuelLabs/sway/issues/5063)
 
 # Summary
 
@@ -14,11 +14,11 @@ eliminate redundant concepts.
 
 [motivation]: #motivation
 
-There currently exists a kludge of features on top of heap pointers that have
+There currently exists a kludge of features on top of pointers that have
 various levels of type correctness, offer various paradigms and have less than
 coherent syntax.
 
-We want to clarify the rules for passing heap data to and from functions, how
+We want to clarify the rules for passing dynamic data to and from functions, how
 they are expressed in the type system and through syntax, and eliminate untyped
 values such as `raw_ptr` and `raw_slice`.
 
@@ -27,7 +27,7 @@ values such as `raw_ptr` and `raw_slice`.
 [guide-level-explanation]: #guide-level-explanation
 
 Unlike Rust, Sway does not manage lifetimes. Due to the nature of smart contract
-execution, heap memory usage only grows and allocated memory is allocated for
+execution, dynamic memory usage only grows and allocated memory is allocated for
 the entire lifetime of the execution. In Rust terms, every reference value has a
 lifetime of `'static`.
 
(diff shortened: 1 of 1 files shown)


rfcs/0010-references.md
```diff
@@ -0,0 +1,423 @@
...
+values such as `raw_ptr` and `raw_slice`.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+Unlike Rust, Sway does not manage lifetimes. Due to the nature of smart contract
+execution, heap memory usage only grows and allocated memory is allocated for
+the entire lifetime of the execution. In Rust terms, every reference value has a
+lifetime of `'static`.
```
Voxelot, line 32: Is this true for references on the stack?

ironcev:
Thanks for formalizing the topic in RFC Jibril! It will be much easier to discuss it now and come to conclusion.
 
> Abstracting away memory management too much might be a drawback for Sway and its positioning as a smart-contract language, however maintaining access to pointer arithmetic should still allow for atypical usage patterns without compromising the use of safe and zero cost abstractions for most cases.
 
IMO the proposal strikes exactly the right balance between the two, let's call them the ML and the Rust world. On the one hand, it removes the burden of explicit memory managment and complex borrowing rules while offering possibilities for compiler optimizations that can result in zero or negligible cost. On the other hand, it leaves the door open for low level C-like pointer arithmetics and access to VM internals, while preserving certain amount of type-safety.
 
That would be my general feedback. I see this as the way to go and I see no major drawbacks, on the contrary.
 
Where I stil see potential for further simplification and clarification is the terminology and partially the proposed syntax.
                                                                                                                                                                                                         
I'll split my detailed feedback into several comments covering these topics:
 
- Pointers
- Slices
- "Stack & Heap" vs "Value and Reference Types" terminology of the abstract memory model
- Rust `Box` vs ML `ref` syntax
- Passing and returning references to and from functions and mutability
- Equality rules

ironcev:
# Pointers

## TL;DR;

For the purpose of this RFC pointers are sketched detailed enough. Can we have a separate RFC where we will specify them in detail and clarify the edge cases, e.g:
 - Available operators like e.g. type-aware ++ and --.
 - Pointers can be passed to functions and returned from functions. Exact syntax etc.
 - The exact result of `__deref(some_reference)`. If it's a reference/pointer, how do we write it as a type?
 - [Potentially] & and * operators.
 - `__ptr_cast`, `*mut ()`, how they play with other operators like ++, ==, etc.

## General feedback

I fully agree with the proposed concept of typed pointers. 

Although it is not explicitly mentioned, I expect the C-like convenience operators like ++ and -- to by type-aware. Is it so? Can we put that explicitly in the proposal?

Out of experience, I also see the need for `__ptr_cast` intrinsic and propose to have it there from the beginning. With 8-bit `u8` and `__prt_cast` library authors can always come to the lowest abstraction possible. I expect these two to practicaly remove the need for `*mut ()`.

Later on, we can thing of offering the the address and dereference operator (& and *) as a syntax sugar for `__addr_of` and `__deref`. Assuming how rarely are pointers supposed to be used, I don't see this as crucial, but our `std::lib` devs might see it differently.

## Pointers and references

What is unclear to me out of the example given is, what `__addr_of(some_reference)` gives back? The pointer to the referenced value or the pointer to the reference?

The way I interpret it, the address of the reference.

In the given example, the nature of `val` is unclear to me as well:

```Rust
let val = Box::new(1u64);
let ptr: *mut Box<u64> = __addr_of(val);
let ptr_val = __deref(ptr);
assert_eq(val, ptr_val);
```

The example with the passing by reference to function shows that accessing the boxed value requires writing `val.val`. (This is another topic which I will address in a different comment.) But so far, a simple question: is `ptr` above the pointer to the value `1u64` or the pointer to the reference `Box<u64>` that points to the value `1u64`.

In other words, are this assert true or false?
`assert_eq(val, 1);`

To generalise the question, if we have a reference to an instance of a reference type, what does `__addr_of(reference)` returns?

```Rust
let s = Struct::new();
let ptr: = _addr_of(s);
```

How I see the example, to come to the value we need to `__deref(__deref(ptr))` or `__deref(__deref(__addr_of(Box::new(1u64))))` which is perfectly fine. I would just like to see it specified explicitly. 

## Pointers and functions

Pointers can be passed to functions and returned from functions. Which again opens the question of the exact syntax.

To avoid going into to much details, how about having an RFC to specify the pointers in detail? It should be straightforward.

For the purpose of this RFC, the provided sketch of pointer behavior is more then fine.

ironcev:
# Slices

## TL; DR;

We should support range syntax from day one, and iterators over slices as soon as we implement iterators.

## General feedback

Fully agree with the proposal, but I see the importance of ranges and the range syntax differently.

Unlike pointers, slices will be (should I say heavily?) used in regular everyday Sway code. They must be first class citizens. I don't see it acceptable to have to use intrinsics to work with slices.

Which calls for having the Range types implemented at the same time with slices. I don't see any reason why this would be particularly difficult.

Once we have iterators, slices should support them as well.

In short, I expect Sway programmers to use range syntax and iterators to comfortably work with slices. They should never need to use neither `__slice` nor `__slice_elem` intrinsics.

(More feedback coming soon :-) Thanks once again for compiling the RFC Jibril!)

IGI-111:
Let me first address the pointer remarks.

---

Pointer operators are not defined as part of this RFC on purpose, except to say that arithmetic should be available on pointer types. I think the exact API we provide will be a different discussion. However I can say that I think we should follow [the Rust example](https://doc.rust-lang.org/std/primitive.pointer.html) here.
That is to say, not have operators between pointers directly available, have `add` and `sub` methods with `u64` offsets that use the size of the pointee, methods to add a given amount of bytes and the ability to drop down to a `u64` to do more complex address math if necessary. 
Tangentially, I also want to remind you that `++` and `--` are available neither in Rust nor Sway, by design.

---
Passing pointers to and from functions is fully defined by the present RFC: it is a pass by value of a primitive type. And the syntax is a natural `fn ident<T>(arg: *mut T) -> *mut T`.

---

Now about `__deref`. I think the signature should look like `fn __deref<T>(arg: *mut T) -> T`. All it does it take a pointer type and dereference it.
Obviously, and perhaps this should make it into the RFC and the documentation, dereferencing an invalid pointer is Undefined Behavior.
It's only intended to work on pointers for now. We may want to fiddle with the intrinsic once we get around to implementing a `*` operator, but I don't think we'll need to.

---

`__addr_of` is a trickier beast and I think I'll need to alter the spec here. I initially didn't feel the need to define it much because it's already an intrinsic. Here's the documentation for it:

>```sway
>__addr_of<T>(val: T) -> raw_ptr
>```
>
>**Description:** Returns the address in memory where `val` is stored.
>
>**Constraints:** `T` is a reference type.

The idea here was to make it return a `*mut T` instead of a `raw_ptr` and extend it to work on primitives. But looking at it now working on both primitives and references creates an ambiguity.

 What we should probably do is maintain its current behavior of only working on reference types, make it return a `*mut` and require boxing a primitive if you want to get a pointer to it.
 
I'll have to correct the examples, but it'd look something like:

```sway
let val: u64 = 1u64;
let ptr: *mut u64 = __addr_of(Box::new(val));
let ptr_val: u64 = __deref(val);
assert_eq(val, ptr_val);

let val: Box<u64> = Box::new(1u64);
let ptr: *mut Box<u64> = __addr_of(Box::new(val));
let ptr_val: Box<u64> = __deref(val);
assert_eq(val, ptr_val);

let val: Box<u64> = Box::new(1u64);
let ptr: *mut u64 = __addr_of(val);
let ptr_val: u64 = __deref(val);
assert_eq(val, Box::new(ptr_val));
```

IGI-111:
Now for slices.

I disagree that we should specify ranges and slicing syntax as part of this RFC. It's a definitely important syntactic sugar, but it is sugar nonetheless and we're already specifying a lot here, much more than I'd like for a single RFC.

The language is still in an unfinished state, and while I agree that making sure the users never have to call an intrinsic in normal use is an important goal for 1.0, it is not a requirement at this stage.

We still make breaking changes to the language on a fairly regular basis and adding forward compatible syntactic sugar at a later date is much less intrusive than that.

xunilrj:
This feels strange:

```rust
let val: u64 = 1u64;
let ptr: *mut u64 = __addr_of(Box::new(val));
let ptr_val: u64 = __deref(val);
assert_eq(val, ptr_val);
```

I would expect it to return `*mut Box<u64>`. `Box<T>` can implement `as_ptr()` without any intrinsic, right?

IGI-111:
> `Box<T>` can implement `as_ptr()` without any intrinsic, right?

It can't. You need an intrinsic to convert a reference into a pointer, and `__addr_of` is that instrinsic.

ironcev:
> I would expect it to return *mut Box<u64>

The examples is actually consistent with the new unambiguous definition. We get the pointer to the value in `Box` means exactly `*mut u64`.

ironcev:
> Let me first address the pointer remarks.

Agree with everything! I can live without ++ and -- as long as the API is convenient and consistent :-) and sticking here to Rust `pointer` API is i agree a natural way to go.

The definition of `__addr_of` is consistent now and also straightforward and compatible with what we have. And it still allows us having pointers to pointers/references (not that i wish anybody will need them :-)) Very good!

ironcev:
> ​I disagree that we should specify ranges and slicing syntax as part of this RFC.

Misunderstanding. I didn't mean to suggest specifying ranges as part of this RFC. It is already broad enough in scope :-) After reading the RFC I got the impression that ranges were in someday-maybe bucket whereas I see them as a must for a feature like slices. 

After reading your comment I say we are on the same page here. First things first but ranges are important.

ironcev:
# "Stack & Heap" vs "Value and Reference Types" terminology of the abstract memory model

Instead of talking about _primitive types and reference types_, proposal is to use [value types and reference types](https://en.wikipedia.org/wiki/Value_type_and_reference_type) as terminology. The terms are well established and self explanatory and eliminate the need of explaining that primitive types have value semantics.

Moreover, sticking to these terms actually eliminate the need for terms _stack and heap_ when explaining the abstract memory model. We simply have _value types and reference types_ with corresponding semantics. _Reference types_ can be dynamically sized and recursive. It is compiler's decision where to allocate the instances of those types for a particular VM. Sway programmer do not need to think about it unless they want to deal with pointers and underlying VMs.

Means, the terms _stack and heap_ exist only in the advanced chapter on pointers, asm, VMs, and in that case we are talking about the actual VM's stack and heap, real and not abstract ones. There we can explain compiler's preferences when allocating instances of value or reference types (value types instances on stack or in registers, reference types instance on heap unless they do not escape, etc. etc.). But as RFC says, these are only hints and preferences.

I am convinced that sticking to this terminology will help Sway to communicate it's memory model in a more clear way that leaves no space for misconceptions like "values are on stack, refs on heap", especially in the environment where we, as the RFC well points it, can target VMs that do not even have one of the two.

esdrubal (approved):

xunilrj (approved):

anton-trunov (approved):
Great read! It should be incorporated into the Sway book, I think

ironcev:
> It should be incorporated into the Sway book, I think

Definitely! These are becoming the fundamentals that Sway programmers will need to understand. Also, adjusting those very few places in the Sway book (I encountered only two) where heap and stack are explicitly mentioned to the value and ref terminology.

PR #33, opened 2023-11-07
Amend References RFC

ironcev:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/amend-references/rfcs/0011-references.md)

The original References RFC (#31) was based on assumption that the types that we internally call _reference types_ (aggregates) also have the reference semantics from the Sway programmers point of view. That is not the case. All our types have value semantics by default (with all the benefits coming from that approach).

Since implementation of references via simple `Box<T>` is not possible, we amend the original proposal with references being a new language concept and a first-class citizen in Sway. This amendment considers only the first part of the original RFC, References. The other two parts, Pointers and Slices, remain the same as in the original.

The amendment comes with detailed examples of the semantics and syntax of the references which can be found in the file [0010-references.sw](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/amend-references/files/0010-references.sw).


files/0010-references.sw
```diff
@@ -0,0 +1,542 @@
...
+    let x = r_a[0] + r_a[1];
+}
+
+
+/// # Embedding references in aggregates
+/// A reference can be a part of an aggregate.
+/// E.g., we can have an array of references.
+struct A {
+    x: u64,
+}
```
IGI-111, line 105: The code that follows doesn't seem to illustrate this point. Is that intended?

files/0010-references.sw
```diff
@@ -0,0 +1,542 @@
...
+/// references.
+///
+/// Let's take a `Vec` as an example. If `Vec` provides access to the elements, only by returning
+/// the copies, the possibility of dangling references is removed by the API design.
+/// But if it provides a method like `get_elem_by_ref(self, index: usize)` we can have a situation where
+/// the caller keeps the reference to an element, calls, e.g., `Vec::clear()`, and by mistake continues
+/// using the reference to a semantically non-existing value.
+///
+/// The question if something can be done on the language level to avoid dangling references
+/// is out of scope od References RFC.
```
IGI-111, line 242:
```suggestion
/// is out of scope of the References RFC.
```

files/0010-references.sw
```diff
@@ -0,0 +1,542 @@
...
+// Ideally, we want to have escape analysis so that in the case of the function below, we can
+// generate a warning for the first argument saying:
+// Copying immutable compound types is expensive and in this case not necessarry. Use `a: &A` instead. 
+fn fn_takes_references(a: A, mut m_a: A, r_a: &A, r_m_a: &mut A, mut m_r_m_a: &mut A) {
+    let x = a.x + m_a.x + r_a.x + m_r_a.x + m_m_r_a.x;
+
+    a.x = 1; // ERROR: `a` is not mutable.
+    m_a.x = 1; // OK. The change does not affect the original (passed) value.
+    
+    r_a.x = 1; // ERROR: `r_a` is a reference to a non muttable `A`.
```
IGI-111, line 290:
```suggestion
    r_a.x = 1; // ERROR: `r_a` is a reference to a non mutable `A`.
```

files/0010-references.sw
```diff
@@ -0,0 +1,542 @@
...
+
+/// # Passing and returning references from functions
+/// References can be passed and return from functions.
+/// It is valid to return a reference to a local value.
+/// The lifetime of the referenced value will automatically be extended by the compiler.
+/// References returned from functions are l-values.
+
+// Ideally, we want to have escape analysis so that in the case of the function below, we can
+// generate a warning for the first argument saying:
+// Copying immutable compound types is expensive and in this case not necessarry. Use `a: &A` instead.
```
IGI-111, line 283:
```suggestion
// Copying immutable compound types is expensive and in this case not necessary. Use `a: &A` instead. 
```

files/0010-references.sw
```diff
@@ -0,0 +1,542 @@
...
+/// # Equality of references
+/// Operator == will compare the referenced content if the referenced type implements `std::ops::Eq`.
+/// To compare references as pointers (checking if they point to the same memory location), the 
+/// `__eq` intrinsic has to be used.
+fn equality_of_references() {
+    let r_a = &2;
+    let r_b = &(1 + 1);
+
+    assert(r_a == r_b);       // Comparing referenced values.
+    assert(!__eg(r_a, r_b));  // Comparing memory locations.
```
IGI-111, line 513: did you mean `__eq`?

ironcev: Yes :-)

IGI-111:
Looks like this page https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/details-of-avx-intrinsics.html bans the github actions robot.

IGI-111 (approved):

ironcev:
> Looks like this page https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/details-of-avx-intrinsics.html bans the github actions robot.

Yes. I've found another page with exactly the same content, essentially a copy of the Intel documentation. We can use that one: http://www.cis.uoguelph.ca/~wgardner/intel_docs/compilers/compiler_c/main_cls/intref_cls/common/intref_avx_details.htm

files/0010-references.sw
```diff
@@ -0,0 +1,666 @@
...
+    a = new_a; // ERROR. `a` is not mutable.
+    m_a = new_a; // OK. Replaces `a` but the change does not affect the original.
+    r_a = &new_a; // ERROR. `r_a` is not mutable.
+    r_m_a = &new_a; // ERROR. `r_m_a` is not mutable.
+    m_r_m_a = &mut new_a; // ERROR: `new_a` is not mutable.
+    m_r_m_a = &mut new_m_a; // OK. Redirecting the mutable reference to a new mutable value.
+    m_r_m_a = &new_m_a; // ERROR: If redirected, `m_r_m_a` must be still a reference to a mutable value.
+}
+
+fn fn_returns_ref() &A { // Reference to immutable A.
```
IGI-111, line 314:
```suggestion
fn fn_returns_ref() -> &A { // Reference to immutable A.
```

files/0010-references.sw
```diff
@@ -0,0 +1,666 @@
...
+    &A { x: 0 } // Returning references to local values is allowed.
+}
+
+fn fn_returns_ref_to_mut() &mut A { // Reference to mutable A.
```
IGI-111, line 318:
```suggestion
fn fn_returns_ref_to_mut() -> &mut A { // Reference to mutable A.
```

sdankel (approved):

jjcnn (changes requested):
I've only got one comment which I think should be reconsidered. Otherwise it looks good.

We should be careful about implicit coersion in pattern matches, e.g., 
```
struct S { a : &[u64] }

x : &S = ...;
match ((y as x).a[0]) {
   < y should be of type &S, even though x gets implicitly dereferenced>
}
```
As I understand it `as` patterns are not currently part of Sway, so it's not something to worry about at this stage.

rfcs/0010-references.md
```diff
@@ -37,62 +37,119 @@ In most cases, you shouldn't need to use explicit pointer types.
...
-let val: Box<u64> = Box::new(1);
-let ptr: *mut u64 = __addr_of(val);
+let ref: &u64 = &1;
+let ptr: *mut u64 = __addr_of(ref); // Returns the address od the referenced value.
 let ptr_val: u64 = __deref(ptr);
-assert_eq(val, Box::new(ptr_val));
+assert_eq(ref, &ptr_val); // Equality of references is the equality of referenced values.
 ```
 
 Dereferencing an invalid pointer is Undefined Behavior.
```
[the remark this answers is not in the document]
xunilrj: But can we guarantee that every deference of invalid pointers will panic? It is interesting to notice that there is no definition of invalid pointers. 

By `Every time the VM runs, a single monolithic memory of size VM_MAX_RAM bytes is allocated` and `The stack begins at 0 and grows upward. The heap begins at VM_MAX_RAM and grows downward.
` (https://specs.fuel.network/master/fuel-vm/index.html)
and the fact that loads only panics when they go beyond VM_MAX_RAM; we have that deferencing invalid pointers  will only revert with overflowing indices and pointer arithmatic. All others cases will read garbage and not revert. 

Is that right?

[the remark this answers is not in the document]
ironcev: @jjcnn @xunilrj Good points and valid concerns. I've added a definition of an invalid pointer and gave an explanation of behavior for FuelVM.

In the case of FuelVM the [access rights](https://specs.fuel.network/master/fuel-vm/index.html#access-rights) will ensure instructions revert in case of a memory access outside of the allocated area. But this is just FuelVM behavior. Once we start targeting other VMs this behavior might be different.

To the general question of runtime checks imposed by the Sway compiler, if a discussion is needed, let's make it as a separate amendment. The above statement on dereferencing invalid pointers was in the accepted part of the original RFC. It wasn't changed by this amendment that focus only on references (`&T`). If there are alternative ideas and concerns, let's rather have and additional amendment that will focus only on pointers. At the offsite last year we were discussing reference tracking and counting and are aware of trade offs between undefined and runtime-checked behavior. Also, note that we have similar tradeoffs in other areas. E.g., we opted for encoding format that does not carry any meta information.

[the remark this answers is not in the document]
ironcev: > I guess it depends whether we can determine the length of arrays.

In case of `[T;n]` arrays we already have compile-time checks for out of bounds access. But of course they work only for const-eval indices.

xunilrj (approved):

vaivaswatha (approved):

PR #39, opened 2024-05-21
Upgradeable contracts using proxies.

JoshuaBatty:
[rendered]

### DIFF
--- a/rfcs/0013-contract-proxies.md
+++ b/rfcs/0013-contract-proxies.md
@@ -1,0 +1,91 @@
+- Feature Name: `contract_proxies`
+- Start Date: 2024-05-21
+- RFC PR: [FuelLabs/sway-rfcs#0013](https://github.com/FuelLabs/sway-rfcs/pull/39)
+
+# Summary
+
+[summary]: #summary
+
+This proposal introduces a feature to simplify the creation of upgradeable contracts in Sway by enhancing the Forc.toml manifest file.
+
+# Motivation
+
+[motivation]: #motivation
+
+Sway developers want to upgrade their smart contract code post-deployment and need tooling support to facilitate this process. They seek a straightforward method, such as adding a configuration to the Forc.toml file, to enable and manage contract upgradeability seamlessly.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+### Use of the Forc.toml for an upgradeable contact.
+This proposal enhances the Forc.toml file to simplify creating upgradeable contracts. A new `[proxy]` table indicates to `forc-deploy` that this contract is upgradable. During the initial deployment, `forc-deploy` stores the proxy contract's address in the `address` field, streamlining future updates and ensuring the address is stored safely with the project.
+
+```toml
+[project]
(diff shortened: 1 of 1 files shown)


rfcs/0013-contract-proxies.md
```diff
@@ -0,0 +1,91 @@
...
+
+```toml
+[project]
+authors = ["Fuel Labs <contact@fuel.sh>"]
+entry = "main.sw"
+license = "Apache-2.0"
+name = "implementation"
+
+[proxy]
+enabled = true # Undecided if we really need this flag yet.
```
kayagokalp, line 32: I think removing this makes sense, as absence or existence of `[proxy]` table can be used to deduce this

kayagokalp (approved):

rfcs/0013-contract-proxies.md
JoshuaBatty: I guess setting this to false would allow you to deploy contracts for testing purposes before generating and committing to the proxy contract.

Users could always comment out `[proxy]` but this might make that more explicit?

kayagokalp: Hmm you mean to test the implementation contract right? if so sounds good to me. Another option to infer the `enabled = true` if the table is there, and for testing users can use `enabled = false` or we can explicitly require the field to be there.

kayagokalp (approved):

rfcs/0013-contract-proxies.md
IGI-111: fwiw I think having them comment it out for that behavior (or run a specific flag) looks more correct.

rfcs/0013-contract-proxies.md
```diff
@@ -0,0 +1,94 @@
...
+
+### **Security Considerations**
+
+- **Address Verification**: Ensure the address stored in **`Forc.toml`** is accurate and not tampered with. Use checksums or other verification methods if needed.
+- **State Management**: Ensure the proxy contract's storage is correctly updated to point to the new implementation contract to prevent any potential issues with state consistency.
+- **Access Control**: Implement proper access control in the proxy contract to restrict who can update the implementation address to prevent unauthorized changes.
+- **Storage Layout Consistency**: Ensure that the storage layout between the proxy and the implementation contracts is identical.
+
+### **Example Contract Syntax**
+@IGI-111 to fill out.
```
sdankel, line 90: @kayagokalp Can we fill this in?

rfcs/0013-contract-proxies.md
alfiedotwtf: I'm not sure if `address = #` is valid TOML.

Some blockchains make things like this explicit by pointing to well-known values like `addresss = 0x0000000000000000000000000000000000000000`. I personally don't like this because you have to count the number of `0` for no good reason and is also error-prone.

Maybe an enum here or a shorter value as a placeholder could make sense:

    address = 0x0
    address = ""

alfiedotwtf: If the proxy will only ever be a single value, would it make sense to move this to the `project` table?

alfiedotwtf (approved):

PR #40, opened 2024-06-10
Configurable and composable storage

ironcev:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/ironcev/storage-v3/rfcs/0013-configurable-and-composable-storage.md)

This RFC introduces a concept of a `Storage` trait, as well as the `StorageBox` and `StorageEncodedBox` structs. Those types are cornerstones upon which we build a fully flexible, feature rich, and robust access to storage. These concepts simplify defining dynamic storage types, like e.g., `StorageVec`. They also provide a dedicated way of configuring (in the `storage` declaration) and initializing (in the code) arbitrary storage types composed of existing storage types.

Additionally, the RFC provides API design guidelines for storage type's APIs. Those guidelines ensure that various aspects of a storage access like, e.g., accessing uninitialized storage, are alway treated in the same way, across various storage types.

### DIFF
--- a/files/0013-configurable-and-composable-storage/api-design/storage_pair.sw
+++ b/files/0013-configurable-and-composable-storage/api-design/storage_pair.sw
@@ -1,0 +1,92 @@
+// This example shows how to apply the proposed API design guidelines
+// on the `StoragePair`.
+//
+// For API design quidelines, see `api-design/storage_vec.sw` and `api-design/storage_encoded_box.sw`.
+//
+// For the complete `Storage` implementation of the `StoragePair` and its other aspects
+// see `sway-libs/storage/storage_pair.sw`.
+//--
+
+impl<A, B> StoragePair<A, B> where A: Storage, B: Storage {
+    const fn get_first_element_self_key(self_key: &StorageKey) -> StorageKey {
+        //--
+        // Returns the self key of the first element of the pair.
+        // `self_key` is the self key of the `StoragePair`.
+        // For the sample implementation, see `sway-libs/user-defined-libs/storage_vec.sw`.
+        //--
+    }
+
+    /// Returns the self key of the second element of the pair.
+    /// `self_key` is the self key of the [StoragePair].
+    const fn get_snd_element_self_key(self_key: &StorageKey) -> StorageKey {
+        //--
+        // Returns the self key of the second element of the pair.
+        // `self_key` is the self key of the `StoragePair`.
+        // For the sample implementation, see `sway-libs/user-defined-libs/storage_vec.sw`.
+        //--
+    }
+
+    //--
+    // The below methods consistently apply the API guidelines
+    // exaplained in detail in `api-design/storage_vec.sw` and `api-design/storage_encoded_box.sw`.
+    //--
+
+    #[storage(read)]
+    pub const fn first(&self) -> A {
+        A::new(Self::get_first_element_self_key(&self.self_key))
+    }
+
+    #[storage(read)]
+    pub const fn snd(&self) -> B {
+        B::new(Self::get_snd_element_self_key(&self.self_key))
+    }
+
+    #[storage(write)]
+    pub fn set_first(&mut self, value: &A::Value) -> A {
+        let first_element_self_key = Self::get_first_element_self_key(&self.self_key);
+        A::init(&first_element_self_key, value)
+    }
+
+    #[storage(write)]
+    pub fn set_first_deep_clear(&mut self, value: &A::Value) -> A where A: DeepClearStorage {
+        let first_element_self_key = Self::get_first_element_self_key(&self.self_key);
+        self.set_element_deep_clear::<A>(&first_element_self_key, value)
+    }
+
+    #[storage(write)]
+    pub fn set_snd(&mut self, value: &B::Value) -> B {
+        let snd_element_self_key = Self::get_snd_element_self_key(&self.self_key);
+        A::init(&snd_element_self_key, value)
+    }
+
+    #[storage(write)]
+    pub fn set_snd_deep_clear(&mut self, value: &B::Value) -> B where B: DeepClearStorage {
+        let snd_element_self_key = Self::get_snd_element_self_key(&self.self_key);
+        self.set_element_deep_clear::<B>(&first_element_self_key, value)
+    }
+
+    #[storage(write)]
+    fn set_element_deep_clear<TElement>(element_self_key: &StorageKey, value: &TElement::Value) -> TElement where TElement: DeepClearStorage {
+        match TElement::internal_layout() {
+            // If the size of the element is fixed, we will just overwrite the current content, if any.
+            ContinuousOfKnownSize(_) => { },
+            // Otherwise, we must first deep clear the content.
+            _ => TElement::new(element_self_key).deep_clear(),
+        }
+
+        TElement::init(element_self_key, value)
+    }
+
+    #[storage(write)]
+    pub fn clear(&mut self) {
+        /* ...
+           Clear only the slot at the `self_key` by calling the `storage::internal::clear::<T>(<key>)`
+           where `T` is a type that guarantees a single slot gets cleared.
+           
+           TODO-DISCUSSION: See the discussion on clearing API in the `internal.sw`.
+           ...
+        */
+    }
+
+    /* ... Other `StoragePair` methods that follow the same API design guidelines. ... */
+}
--- a/files/0013-configurable-and-composable-storage/sway-libs/storage.sw
+++ b/files/0013-configurable-and-composable-storage/sway-libs/storage.sw
@@ -1,0 +1,245 @@
+//--
+// Unlike the current `StorageKey`, the new one becomes a plain struct that holds
+// the `slot` and the `offset`. There will be no storage type impls on it.
+//--
+pub struct StorageKey {
+    slot: b256,
+    offset: u64,
+}
+
+impl StorageKey {
+    pub const fn new(slot: b256, offset: u64) -> Self {
+        Self {
+            slot,
+            offset
+        }
+    }
+}
+
+/// Defines how a concrete [Storage] implementation places
+/// its stored values within the storage, relative to the
+/// given self key of the storage.
+pub enum StorageLayout {
+    /// The values are scattered within the storage and
+    /// can be on arbitrary storage slots, unrelated to
+    /// the self key.
+    Scattered: (),
+    /// The values are stored starting from the self key
+    /// in a continuous sequence of consecutive storage
+    /// slots. The size that the stored value occupies is
+    /// not know and can vary.
+    Continuous: (),
+    /// The values are stored starting from the self key
+    /// in a continuous sequence of consecutive storage
+    /// slots. The size that the stored value occupies is
+    /// known and fixed.
+    ContinuousOfKnownSize: u64,
+}
+
+//--
+// Functions and methods marked with `internal` are used only by Storage
+// implementors and by the compiler.
+//
+// TODO-DISCUSSION: Should we introduce a language feature here and not rely on a naming convention hack?
+//                  E.g., a language feature for something like `pub(impl) fn`.
+//                  The naming convention solution is convenient and simple, but still bloats the list
+//                  of methods/functions availabe in a scope.
+//--
+pub trait Storage {
+    /// The type of the value that can be stored in this [Storage].
+    type Value;
+    /// The type that describes how this [Storage] stores its internal data and the
+    /// stored value within the storage.
+    ///
+    /// This type can be only one of the valid combinations of [StorageConfig<TValue>]s described below.
+    ///
+    /// Sway compiler will use this type to properly configure storage slots for the elements declared
+    /// in `storage` declarations.
+    ///
+    /// `Config` type is defined recursively as:
+    ///
+    /// ```ignore
+    /// <Config> := StorageConfig<TValue>
+    ///             | [<Config>]
+    ///             | (<Config>, ...);
+    /// ```
+    ///
+    /// Where `[<Config>]` represents a slice of `Config`s and `(<Config>, ...)` a tuple of arbitrary
+    /// many `Config`s.
+    type Config;
+
+    /// Creates a new [Storage] that is not initialized.
+    ///
+    /// To create a new initialized [Storage], use the [Storage::init] constructor.
+    //--
+    //  Compiler will call this function in the storage element access.
+    //  E.g. in `storage.x` to create `x`.
+    //
+    //  Note that `new` is a `const` function. This allows us to use storage
+    //  elements when configuring other storage elements, e.g., to obtain
+    //  storage references within storage configuration. For an example,
+    //  see `demo_contract_storage_references.sw`.
+    //--
+    const fn new(self_key: &StorageKey) -> Self;
+
+    /// Creates a configuration information for configuring an instance of this [Storage].
+    /// The [Storage] will be located at the `self_key` and has to store `values`.
+    ///
+    /// This function should be used only when developing a custom
+    /// [Storage] and should never occur in contract code.
+    //--
+    //  Compiler will call this function when generating storage slots.
+    //  It will pass the `self_key` that it has generated for the storage element
+    //  and the `value` will be the the const evaluated configuration value
+    //  defined at the RHS of the `:=` operator.
+    //--
+    const fn internal_get_config(self_key: &StorageKey, value: &Self::Value) -> Self::Config;
+
+    /// The [StorageLayout] of this [Storage].
+    ///
+    /// This function should be used only when developing a custom
+    /// [Storage] and should never occur in contract code.
+    const fn internal_layout() -> StorageLayout;
+
+    /// The self key of this [Storage].
+    const fn self_key(&self) -> StorageKey;
+
+    /// Creates a new [Storage] that is initialized to the `value`.
+    /// The created [Storage] is located at the `self_key` and stores the `value`.
+    //--
+    // TODO-DISCUSSION: Should we pass the `&Self::Value` here or just the `Self::Value`?
+    //                  This is a general question that is not only related to the storage API.
+    //                  It will become relevant when we introduce references in the STD.
+    //                  It is about avoiding copying data but not having the move semantics.
+    //                  How much can we optimize? How to avoid heap allocations in case of
+    //                  using references? Etc.
+    //--
+    #[storage(write)]
+    fn init(self_key: &StorageKey, value: &Self::Value) -> Self;
+} {
+    const fn as_ref(&self) -> StorageRef<Self> {
+        StorageRef::<Self>::Ref(self.self_key())
+    }
+}
+
+/// Provides information to the compiler, during the configuration of the `storage`,
+/// at which `storage_key` to store the `value`.
+///
+/// `TValue` cannot be a [Storage] or a [StorageConfig].
+///
+/// This struct should be used only when developing a custom
+/// [Storage] and should never occur in contract code.
+pub struct StorageConfig<TValue>
+{
+    pub storage_key: StorageKey,
+    pub value: TValue,
+}
+
+/// Stores a reference to a [Storage] of type `TStorage`.
+///
+/// To create a [StorageRef::Ref] use [Storage::as_ref].
+///
+/// To create a null reference, use [StorageRef::Null].
+///
+/// Internally, a [StorageRef] stores the storage key
+/// of the referenced storage entity.
+///
+/// [StorageRef] requires two storage slots to store a reference.
+/// Consider using other, less storage consuming, approaches to "reference"
+/// storage entities. E.g., if they are stored in a [StorageVec], consider
+/// "referencing" them by storing their indices as "references".
+//--
+// This type is a type safe wrapper around `StorageKey` that
+// ensures we are always referencing and dereferencing a
+// proper `TStorage` type.
+//--
+pub enum StorageRef<TStorage> where TStorage: Storage {
+    Null: (),
+    Ref: StorageKey,
+}
+
+impl<TStorage> StorageRef<TStorage> where TStorage: Storage {
+    /// Reverts if the [StorageRef] is [StorageRef::Null].
+    fn deref(&self) -> TStorage {
+        self.try_deref().unwrap()
+    }
+
+    const fn try_deref(&self) -> Option<TStorage> {
+        match self {
+            Self::Null => None,
+            Self::Ref(storage_key) => Some(TStorage::new(storage_key)),
+        }
+    }
+
+    const fn is_null(&self) -> bool {
+        match self {
+            Self::Null => true,
+            _ => false,
+        }
+    }
+}
+
+/// A [Storage] that supports reading and retrieving its entire stored value.
+pub trait DeepReadStorage: Storage {
+    /// Returns the entired value stored in the [Storage],
+    /// or `None` if the [Storage] is uninitialized.
+    ///
+    /// Composable storage types usually offer specialized
+    /// methods for accessing parts of the stored value.
+    /// Retrieving the entire stored value should be used
+    /// only if the entire value is actually needed.
+    ///
+    /// Not all storage types can retrieve the value
+    /// they store. E.g., a [StorageMap] is not aware of
+    /// all the elements stored and require access by
+    /// key to individual elements.
+    ///
+    /// Retrieving the entire stored value can require
+    /// high amout of storage reads for certain storage
+    /// types. Therefore, this method must always be
+    /// used with care and only if the entire value is
+    /// actually needed.
+    #[storage(read)]
+    fn try_deep_read(&self) -> Option<Self::Value>;
+} {
+    #[storage(read)]
+    fn deep_read(&self) -> Self::Value {
+        self.try_deep_read().unwrap()
+    }
+}
+
+/// A [Storage] that supports removing its entire stored value
+/// from the storage by uninitializing the storage slots which
+/// were occupied by the stored value.
+pub trait DeepClearStorage: Storage {
+    /// Clears the entired value stored in the [Storage].
+    ///
+    /// This call always succeeds, even if the [Storage] is
+    /// uninitialized.
+    ///
+    /// Composable storage types usually offer specialized
+    /// methods for semantically clear the stored value.
+    /// To clear semantically means that the value will be removed
+    /// from the perspective of the storage type, but the
+    /// occupied storage slots will not necessarily be uninitialized.
+    ///
+    /// E.g., semantically clearing a [StorageVec] could simply set the length
+    /// of the vector to zero, without actually touching the stored
+    /// elements.
+    ///
+    /// Clearing the entire stored value should be used
+    /// only if clearing the entire value is actually needed.
+    ///
+    /// Not all storage types can clear the value
+    /// they store. E.g., a [StorageMap] is not aware of
+    /// all the elements stored and therefore cannot clear
+    /// all of them.
+    ///
+    /// Clearing the entire stored value can require
+    /// high amout of storage clears for certain storage
+    /// types. Therefore, this method must always be
+    /// used with care and only if clearing the entire value is
+    /// actually needed.
+    #[storage(write)]
+    fn deep_clear(&mut self);
+}
--- a/rfcs/0013-configurable-and-composable-storage.md
+++ b/rfcs/0013-configurable-and-composable-storage.md
@@ -1,0 +1,600 @@
+ Name- Feature: `configurable_and_composable_storage`
+- Start Date: 2024-06-10
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/001)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+
+# Summary
+
+[summary]: #summary
+
+This RFC introduces a concept of a `Storage` trait, as well as the `StorageBox` and `StorageEncodedBox` structs. Those types are cornerstones upon which we build a fully flexible, feature rich, and robust access to storage. These concepts simplify defining dynamic storage types, like e.g., `StorageVec`. They also provide a dedicated way of configuring (in the `storage` declaration) and initializing (in the code) arbitrary storage types composed of existing storage types.
+
+Additionally, the RFC provides API design guidelines for storage type's APIs. Those guidelines ensure that various aspects of a storage access like, e.g., accessing uninitialized storage, are alway treated in the same way, across various storage types.
+
+# Motivation
+
+[motivation]: #motivation
+
+The [`storage_keys` RFC](./0008-storage-handler.md) introduced the concept of a `StorageKey` which improved the creation of dynamic storage types, like `StorageVec` and `StorageMap`. While bringing noticeable improvement at the time, that approach, which we currently use, has the following issues and limitations:
+- There is no clear definition of a "storage type", or a "storable type".
+  - E.g., the `x: u64 = 123` syntax only gives a (false) impression that we are storing a `u64` in the storage, while actually just being a syntax sugar for `x: StorageKey<u64> = <compiler-generated configuration of storage slots based on encoded u64>`.
+  - This is especially confusing for dynamic storage types, because they are always, from the Sway perspective, just empty structs. Their real definition is an impl of `StorageKey<T>`. E.g., `StorageVec<V>` becomes `impl<V> StorageKey<StorageVec<V>>`.
+  - Above points lead to confusing error messages when using `storage` and forbid having methods like `clear()` or `read()` on dynamic storage types, because they clash with already existing methods implemented for every `StorageKey`.
+  - Similarly, the common methods, like the mentioned `clear()` and `read()` might not have sense for certain storage types but they are still always available.
+- Composing dynamic storage types, e.g., `StorageVec<StorageVec<u64>>`, relays on the `field_id` hack in the `StorageKey`.
+  - `field_id` represents an unnecessary cognitive burden when using `StorageKey`s.
+  - Proper usage of `filed_id` is error-prone and its false usage, which is easy to happen, can lead to hard-to-explain bugs like, e.g., [this one](https://github.com/FuelLabs/sway/issues/6036F).   
+- Clearing dynamic storage types, e.g., `StorageVec<T>`, relays on a global hack in the `clear()` method of the `StorageKey`.
+- Dynamic storage types cannot be configured in the `storage` declaration.
+  - Those types are always configured as empty Sway structs. E.g., `StorageVec { }`.
+  - Programmers are forced to alway write those empty structs in the configuration, just to satisfy the compiler.
+  - The only way to configure dynamic storage types is via SDKs, or by using them in code, thus at runtime.
+- As a consequence of the above points, but also a separate requirement, it is not possible to write a const eval function that could configure a dynamic storage type based on a certain logic.
+- As a consequence of the above points, but also a separate requirement, it is not possible to configure a dynamic storage type composed of other storage types, like e.g., `StorageMap<u64, StorageVec<u256>>`.
+- Elements in the `storage` declaration must alway be configured, even if programmers just want to list them and actually configure them during deployment via SDKs.
+- It is not possible to store dynamic Sway types like `Vec` and `String` in the storage.
+
+The above points lead to:
+- non-optimal developer experience (e.g., missing possibility to configure dynamic storage types in `storage` declarations; being forced to configure all `storage` elements, etc.).
+- difficult and error-prone development of user-defined dynamic storage types due to error-prone storage access design and storage API.
+- possibility to provide inconsistent APIs (e.g., `StorageVec::pop()` returns `V` while `StorageVec::get()` returns `StorageKey<V>`).
+
+With the approach proposed in this RFC, we will solve all the above points by providing:
+- clear abstractions for storable types via `Storage` trait, as well as storage primitives for storing arbitrary Sway types, even the dynamic ones.
+  - The abstractions will forbid semantically wrong usage by-design.
+- support for safe and arbitrary composable and configurable dynamic storage types.
+  - Implementing the `Storage` trait will automatically make a storage type compatible with all other storage types.
+- support for complex configuration of dynamic storage types in `storage` declarations.
+- support for skipping default configuration of elements in `storage` declarations.
+
+## Sample code
+
+[sample-code]: #sample-code
+
+To prove that such implementation is possible and to give a tangible feeling for the proposed new approach to storage handling, this RFC comes with extensive [sample code](../files/0013-configurable-and-composable-storage/):
+- [sway-libs](../files/0013-configurable-and-composable-storage/sway-libs/) provide:
+  - definition of the `Storage` trait as well as few other traits like, e.g., `DeepReadStorage` and `DeepClearStorage`.
+  - implementation of the atomic `Storage` types, `StorageBox` and `StorageEncodedBox`.
+  - implementation of STD dynamic storage types like, e.g., `StorageVec` and `StorageMap`.
+- [user-defined-libs](../files/0013-configurable-and-composable-storage/user-defined-libs/) provide:
+  - implementation of a user defined `StoragePair` storage type.
+- [contracts](../files/0013-configurable-and-composable-storage/contracts/) provide:
+  - several demo-contracts that demonstrate usage of the new storage.
+- [api-design](../files/0013-configurable-and-composable-storage/api-design/) provides:
+  - implementation of several storage types with the focus on explaining the proposed API design guidelines.
+
+The sample code contains documentation for Sway programmers (using Sway doc-comments `///`), as well as extensive explanations for RFC reviewers (using comments starting with `//--`). Open discussion points are marked with `TODO-DISCUSSION`.
+
+Going through the sample code is highly encouraged.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+This guide-level explanation primarily explains how the new `storage` declaration and `Storage` types will be used by Sway programmers. Explaining writing user-defined `Storage` types was more fitting to the [Reference-level explanation](#reference-level-explanation) and is explained there.
+
+## `storage` declarations and storage types
+
+[storage-declarations-and-storage-types]: #storage-declarations-and-storage-types
+
+The `storage` declaration will get a new operator, `:=`, pronounced _configured with_. The left-hand side (LHS) of the `:=` operator will be the definition of the storage element, as it is now. The right-hand side (RHS) will represent the _configuration_ of that storage element. The configuration is any const eval expression that returns an instance of a type stored by the storage type specified on the LHS.
+
+The type of the LHS storage element must implement the `Storage` trait. This trait denotes a storage type that can be an element of a `storage` declaration. The `Storage` trait is explained in detail in the [The `Storage` trait](#the-storage-trait) chapter of the [Reference-level explanation](#reference-level-explanation).
+
+There are two classes of storage types. _Atomic storage types_ are the leafs of hierarchies of composed storage types. They cannot contain other storage types. _Compound storage types_ are storage types that contain other storage types. Standard library (STD) provides two _atomic storage types_, the [`StorageBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_box.sw) and the [`StorageEncodedBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw), as well as several compound storage types, like e.g., [`StorageVec`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_vec.sw) and [`StorageMap`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_map.sw).
+
+To store Sway types of a known size that does not contain pointers or references, the `StorageBox` is used. To store a Sway type of a dynamic size, that implements `AbiEncode` and `AbiDecode`, the `StorageEncodedBox` is used.
+
+```Sway
+storage {
+    box_1: StorageBox<u64> := 0,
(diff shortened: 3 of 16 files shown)


rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+1. Certain operations might have a version with a "special behavior". E.g., in the case of the `StorageVec`, we want to provide a possibility to `get()` an element without the expensive length check. In such cases, the method will be named by the original name, and the suffix indicating the "special behavior". E.g., `StorageVec::get_unchecked()`.
+1. When the special behavior consists of doing deep clear via `DeepClearStorage` or deep read via `DeepReadStorage`, the suffix will always be `_deep_clear` and `_deep_read`, respectively.
+
+Thus, every storage type operation might come in all or some of these flavours:
+- `fn <operation>() -> T`: reverts in case of uninitialized storage or other error.
+- `fn try_<operation>() -> Option<T>`
+- `fn <operation>_<special_behavior>() -> T`: reverts in case of uninitialized storage or other error.
+- `fn try_<operation>_<special_behavior>() -> Option<T>`
```
IGI-111, line 287: We discussed changing the naming in the API previously here: https://github.com/FuelLabs/sway/issues/4740

I think we should go with the opposite of what's said in this paragraph.

Most of the time, you want an operation to return `Option`. I'd much rather we standardize on `read`+`read_unchecked` than `try_read` + `read`. Reverts should be the exception not the rule and it should be clear when an operation potentially reverts.

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+        StorageRef::Null,
+        storage.vec_1.as_ref(), // <<-- Note using the `storage` keyword here, as well as `as_ref`.
+        storage.pair_1.first().as_ref(),
+        storage.map_1.get("abc").as_ref(),
+        StorageRef::Ref(some_const_fn_that_returns_the_storage_key_of_the_referenced_storage()),
+    ],
+}
+```
+
+For more examples on using storage references, see the [`demo_contract_storage_references.sw`](../files/0013-configurable-and-composable-storage/contracts/demo_contract_storage_references.sw).
```
IGI-111, line 347: I think `StorageRef` is an unnecessary complication.

I don't see what this adds over `Option<StorageKey>` which would already be supported. If anything it is worse because it reintroduces a potential implicit null.

Unless there's a compelling reason to have a special type, I say we should remove this and recommend storing `StorageKey`s and `Option<StorageKey>`s for references to other parts of the storage.

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+E.g., the `StorageVec<TStorage>` must be able to store the vector's length and the configuration of all the contained elements. Thus, its `Config` can be specified as `type Config = (StorageConfig<u64>, [TStorage::Config])`. The first part of the tuple is the length, and the second the slice of all the `Config`s of the contained storage type.
+
+In general, the `Config` type is defined recursively as:
+
+```
+<Config> := StorageConfig<TValue>
+            | [<Config>]
+            | (<Config>, ...);
+```
```
IGI-111, line 470: The mechanism is sound, but we need better naming than "config". It should be clearer that `StorageConfig` represents a single address initializer for consecutive slots and that `Config` is a whole, potentially composed, initializer.

Maybe `StorageInit` and `AddressInit`?

files/0013-configurable-and-composable-storage/sway-libs/storage.sw
```diff
@@ -0,0 +1,245 @@
...
+
+    /// The self key of this [Storage].
+    const fn self_key(&self) -> StorageKey;
+
+    /// Creates a new [Storage] that is initialized to the `value`.
+    /// The created [Storage] is located at the `self_key` and stores the `value`.
+    //--
+    // TODO-DISCUSSION: Should we pass the `&Self::Value` here or just the `Self::Value`?
+    //                  This is a general question that is not only related to the storage API.
+    //                  It will become relevant when we introduce references in the STD.
```
IGI-111, line 112:
```suggestion
    //                  It will become relevant when we introduce references in the standard library.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+## Sample code
+
+[sample-code]: #sample-code
+
+To prove that such implementation is possible and to give a tangible feeling for the proposed new approach to storage handling, this RFC comes with extensive [sample code](../files/0013-configurable-and-composable-storage/):
+- [sway-libs](../files/0013-configurable-and-composable-storage/sway-libs/) provide:
+  - definition of the `Storage` trait as well as few other traits like, e.g., `DeepReadStorage` and `DeepClearStorage`.
+  - implementation of the atomic `Storage` types, `StorageBox` and `StorageEncodedBox`.
+  - implementation of STD dynamic storage types like, e.g., `StorageVec` and `StorageMap`.
```
IGI-111, line 58:
```suggestion
  - implementation of standard library dynamic storage types like, e.g., `StorageVec` and `StorageMap`.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+The type of the LHS storage element must implement the `Storage` trait. This trait denotes a storage type that can be an element of a `storage` declaration. The `Storage` trait is explained in detail in the [The `Storage` trait](#the-storage-trait) chapter of the [Reference-level explanation](#reference-level-explanation).
+
+There are two classes of storage types. _Atomic storage types_ are the leafs of hierarchies of composed storage types. They cannot contain other storage types. _Compound storage types_ are storage types that contain other storage types. Standard library (STD) provides two _atomic storage types_, the [`StorageBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_box.sw) and the [`StorageEncodedBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw), as well as several compound storage types, like e.g., [`StorageVec`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_vec.sw) and [`StorageMap`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_map.sw).
```
IGI-111, line 84:
```suggestion
There are two classes of storage types. _Atomic storage types_ are the leafs of hierarchies of composed storage types. They cannot contain other storage types. _Compound storage types_ are storage types that contain other storage types. The standard library provides two _atomic storage types_, the [`StorageBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_box.sw) and the [`StorageEncodedBox`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw), as well as several compound storage types, like e.g., [`StorageVec`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_vec.sw) and [`StorageMap`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_map.sw).
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+For more examples of using storage types in storage declarations, see:
+- [`demo_contract_basic_storage.sw`](../files/0013-configurable-and-composable-storage/contracts/demo_contract_basic_storage.sw)
+- [`demo_contract_encoded_storage.sw`](../files/0013-configurable-and-composable-storage/contracts/demo_contract_encoded_storage.sw).
+- [`demo_contract_user_defined_storage.sw`](../files/0013-configurable-and-composable-storage/contracts/demo_contract_user_defined_storage.sw).
+
+## Default `storage` configuration
+
+[default-storage-configuration]: #default-storage-configuration
+
+If developers just want to list the `storage` elements, knowing that they will be configured via SDKs, they can list the elements without providing the _configure with_ operator. The types of the values used to configure the elements must implement the `core::default::Default` trait that will be added to the STD.
```
IGI-111, line 188:
```suggestion
If developers just want to list the `storage` elements, knowing that they will be configured via SDKs, they can list the elements without providing the _configure with_ operator. The types of the values used to configure the elements must implement the `core::default::Default` trait that will be added to the standard library.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+For more examples on using storage references, see the [`demo_contract_storage_references.sw`](../files/0013-configurable-and-composable-storage/contracts/demo_contract_storage_references.sw).
+
+## Breaking changes
+
+[breaking-changes]: #breaking-changes
+
+This proposal brings breaking changes in:
+- `storage` declarations.
+- existing STD storage types like, e.g., `StorageVec`.
```
IGI-111, line 355:
```suggestion
- existing standard library storage types like, e.g., `StorageVec`.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+- utilize [expressive diagnostics](./0012-expressive-diagnostics.md) to provide detailed explanations on how to fix compiler errors and migrate the code.
+- where applicable, provide scripts to semi-automate the migration by converting the existing code to the new syntax and semantics.
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+The implementation of the proposed features relies on a couple of other language features that need to be implemented first. We will only list them here, since their detailed explanation, in certain cases, could require a separate RFC. To follow the reference-level explanation of the configurable and composable storage, it is sufficient to have an intuitive understanding of these language features. Where needed, they are explained in enough detail in the [sample code](../files/0013-configurable-and-composable-storage/).
+
+The [sample code](../files/0013-configurable-and-composable-storage/) provides a detailed view on how the Sway STD part of this RFC will be implemented and used.
```
IGI-111, line 371:
```suggestion
The [sample code](../files/0013-configurable-and-composable-storage/) provides a detailed view on how the standard library part of this RFC will be implemented and used.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+}
+```
+
+The functions prefixed with `internal` are used only when implementing storage types, and should never be used in contract code.
+
+`Storage` trait methods and associated types provide all the information needed by the compiler to:
+- support configuring arbitrary storage types in `storage` declarations.
+- support type-checking arbitrary composition of storage types.
+
+The get a hands-on feeling for concrete implementations of the `Storage` trait and better understanding of the meaning of its methods and associated types see the [provided implementations of STD storage types](../files/0013-configurable-and-composable-storage/sway-libs/storage/) and the user-defined [`StoragePair`](../files/0013-configurable-and-composable-storage/user-defined-libs/storage_pair.sw).
```
IGI-111, line 416:
```suggestion
The get a hands-on feeling for concrete implementations of the `Storage` trait and better understanding of the meaning of its methods and associated types see the [provided implementations of standard library storage types](../files/0013-configurable-and-composable-storage/sway-libs/storage/) and the user-defined [`StoragePair`](../files/0013-configurable-and-composable-storage/user-defined-libs/storage_pair.sw).
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+To get the feeling for these straightforward and essentially short and compact implementations of storage types, see the atomic and compound storage types implemented in [`sway-libs`](../files/0013-configurable-and-composable-storage/sway-libs/) and the user-defined [`StoragePair`](../files/0013-configurable-and-composable-storage/user-defined-libs/storage_pair.sw).
+
+### Defining storage traits
+
+[defining-storage-traits]: #defining-storage-traits
+
+The clear definition of the `Storage` trait allows defining storage traits. A storage trait is a trait that has `Storage` as a supertrait. The STD will come with (at least) two storage traits, `DeepReadStorage` and `DeepClearStorage` both defined in the [`storage.sw`](../files/0013-configurable-and-composable-storage/sway-libs/storage.sw).
```
IGI-111, line 540:
```suggestion
The clear definition of the `Storage` trait allows defining storage traits. A storage trait is a trait that has `Storage` as a supertrait. The standard library will come with (at least) two storage traits, `DeepReadStorage` and `DeepClearStorage` both defined in the [`storage.sw`](../files/0013-configurable-and-composable-storage/sway-libs/storage.sw).
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+To get the feeling for these straightforward and essentially short and compact implementations of storage traits, see the implementations of `DeepReadStorage` and `DeepClearStorage` for [`StorageVec`](../files/0013-configurable-and-composable-storage/sway-libs/storage/storage_vec.sw) and [`StoragePair`](../files/0013-configurable-and-composable-storage/user-defined-libs/storage_pair.sw).
+
+By convention, all storage trait will have the suffix `Storage`.
+
+## Internal API and the `__state` intrinsics
+
+[internal-api-and-the-state-intrinsics]: #internal-api-and-the-state-intrinsics
+
+The STD will still provide the low-level storage `read`, `write`, and `clear` functions. However, unlike now when they are actually sometimes needed in contract code, e.g., to store arrays, in the new storage implementation there should never be a need to use them in contracts. Instead, the atomic `StorageBox` and `StorageEncodedBox` should be used. They provide the same low-level functionality while offering safe storage access.
```
IGI-111, line 552:
```suggestion
The standard library will still provide the low-level storage `read`, `write`, and `clear` functions. However, unlike now when they are actually sometimes needed in contract code, e.g., to store arrays, in the new storage implementation there should never be a need to use them in contracts. Instead, the atomic `StorageBox` and `StorageEncodedBox` should be used. They provide the same low-level functionality while offering safe storage access.
```

IGI-111 (commented):
Capitalized STD is confusing because the standard library is already made up of `std` and `core`, either use those specifically or just say "the standard library".

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+The low-level API will be used only when implementing storage types, and even in those cases not always. Thus, the proposal is to move them to the module named [`internal`](../files/0013-configurable-and-composable-storage/sway-libs/storage/internal.sw) to emphasize that they are, similar to `Storage::internal_` functions, meant to be used only when developing custom storage types.
```
IGI-111, line 554: Direct programmatic manipulation of storage should not be encouraged but it should not be this discouraged either. There are cases where it's a better more straightforward pattern to use to implement what you want.

I do think having it in a separate module is fine, but I don't like the implication that it is "internal", it's merly raw or low level. But it's perfectly valid for the end user and not specific to storage type implementors.

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+Similarly, the `__state_` intrinsics should be used only in the implementations of the internal API functions. In addition, the `__state_load_word` intrinsic should be changed to return the information if the requested slot was set or unset.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+The only drawback I can think of is the time and effort needed to implement the proposal and to deal with the [breaking changes](#breaking-changes). However, the impact of _not_ improving the current storage handling will over time surely be higher. It would mean carrying on the issues mentioned in the [Motivation](#motivation) and, worst of all, living with an API that is error-prone by-design.
```
IGI-111, line 562: Maybe we should have some discussion of the performance implications of this change. Can storage get slower as a result of this, if employed correctly?

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+During the work on this RFC, five different approaches were intensely discussed and modeled in code. Non of them came even near to the simplicity of use and implementation of the proposed approach. Also, all of the discussed approaches could not guarantee an API that would prevent error-prone implementation of storage types, when it comes to the containment and composition semantics.
```
IGI-111, line 568:
```suggestion
During the work on this RFC, five different approaches were intensely discussed and modeled in code. None of them came even near to the simplicity of use and implementation of the proposed approach. Also, all of the discussed approaches could not guarantee an API that would prevent error-prone implementation of storage types, when it comes to the containment and composition semantics.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+- additionally distinguish between operations failing because of the storage being uninitialized and because of semantic errors like out of bound access. 
+- avoid having possible multiple versions of a method by, e.g., putting the `Storage` instance in a "state" where it behaves differently for particular operations.
+
+All three approaches ended up to be less intuitive to use and implement then the approach proposed in the [Storage API design guidelines](#storage-api-design-guidelines).
+
+# Prior art
+
+[prior-art]: #prior-art
+
+As the [`storage_keys` RFC](./0008-storage-handler.md) puts it, there is no much prior art that would correspond with the overall direction for storage handling that we took in Sway. This means, having the `storage` elements be distinguished and having the API that communicates that storage operations might fail.
```
IGI-111, line 580:
```suggestion
As the [`storage_keys` RFC](./0008-storage-handler.md) puts it, there is not much prior art that would correspond with the overall direction for storage handling that we took in Sway. This means, having the `storage` elements be distinguished and having the API that communicates that storage operations might fail.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+
+The proposed approach builds on top of those two premisses, and brings robust and easy to implement composable storage types that can be arbitrarily configured in `storage` declarations.
```
IGI-111, line 582:
```suggestion
The proposed approach builds on top of those two premises, and brings robust and easy to implement composable storage types that can be arbitrarily configured in `storage` declarations.
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+Through the RFC process I expect the following open questions to be resolved:
+- questions posed in the [sample code](../files/0013-configurable-and-composable-storage/) and marked with `TODO-DISCUSSION`.
+- naming of proposed code elements like, e.g., method/function names, type names, etc. Since those names will become the part of the storage API it is of highest importance to come up with good and expressive names for abstractions.
+
+Also, I expect the current storage attributes, `#[storage(read, write)]`, to be discussed. In the sample code, the existing attributes are used. However, there are two questions to be decided on:
+- if the meaning of `write` remains "read and write" as it is now, the attributes should be renamed to `readonly` and `readwrite`.
```
IGI-111, line 593: There are important reasons (for standards, notably) that we cannot be explicitly bound to read and write and must have read only/read write semantics.

Maybe we can just have the smallest change:
```sway
#[storage(read)]
#[storage(read_write)]
```

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+The `Storage` trait is [defined](../files/0013-configurable-and-composable-storage/sway-libs/storage.sw) as:
+
+```Sway
+pub trait Storage {
+    type Value;
+    type Config;
+
+    const fn new(self_key: &StorageKey) -> Self;
+
+    const fn internal_get_config(self_key: &StorageKey, value: &Self::Value) -> Self::Config;
```
IGI-111, line 395: If we go with `StorageInit` instead of `Config`, maybe this can be `internal_make_initializer`

rfcs/0013-configurable-and-composable-storage.md
```diff
@@ -0,0 +1,600 @@
...
+Also, every storage type will always have the same way of configuring, defined by the `Value` type. This guarantees intuitive and consistent usage. E.g., for the `StorageVec`, developers will know that they always need to provide a slice of the values expected by the contained storage type.
+
+The provided RHS expression that returns the type specified by `Value` must be const eval, as it is now.
+
+In the second step, the compiler needs to calculate the concrete storage slots in which to store the _configured with_ RHS values.
+
+This is where the `Storage::Config` associated type comes in play, together with the `const fn` function `internal_get_config()`.
+
+Before generating the slots, the compiler already knows:
+- the `StorageKey` at which a particular `storage` element will be stored. This key is calculated based on the element name and namespace, or it is given using the `in` keyword. This storage key is, as mentioned below, called the _self key_ of the storage element.
```
IGI-111, line 445: The name of "self key" is an artifact of our thought process and it's littered all over this proposal, but I think we should try to come up with a better one since it's no longer involved with `self` as a keyword.

Maybe "field key"? Since it's the `StorageKey` for a given `storage` field?

files/0013-configurable-and-composable-storage/api-design/storage_pair.sw
```diff
@@ -0,0 +1,92 @@
...
+        //--
+        // Returns the self key of the second element of the pair.
+        // `self_key` is the self key of the `StoragePair`.
+        // For the sample implementation, see `sway-libs/user-defined-libs/storage_vec.sw`.
+        //--
+    }
+
+    //--
+    // The below methods consistently apply the API guidelines
+    // exaplained in detail in `api-design/storage_vec.sw` and `api-design/storage_encoded_box.sw`.
```
bitzoic, line 31:
```suggestion
    // explained in detail in `api-design/storage_vec.sw` and `api-design/storage_encoded_box.sw`.
```

bitzoic (commented):
On first look, I am unsure how I feel about including `StorageBox` in the storage definition from a user perspective. It makes sense however with the current storage implementation the `StorageKey` is behind the scenes and makes it much more readable:
```sway
storage {
     box_1: StorageBox<u64> := 0,
}
```
vs 
```sway
storage {
     box_1: u64 = 0, // This is actually a StorageKey<u64>
}
```
It is unclear for me whether this is still achievable on the compiler side.

As `AbiEncode` is [implemented for primitive types](https://github.com/FuelLabs/sway/blob/5e24673b827837690b41dfbffd2bf746d936b90d/sway-lib-core/src/codec.sw#L179) such as `u64`, couldn't `StorageEncodedBox` be used? How do we know we are using using "the correct one"?


I am also wondering what the difference is [here](https://github.com/FuelLabs/sway-rfcs/blob/39c0de269de5dce58a39704438b3bcf5b3e50c7f/files/0013-configurable-and-composable-storage/contracts/demo_contract_encoded_storage.sw#L50C5-L50C138) between `StorageVec` and `Vec`:
```sway
vec_of_encoded_val_1: StorageVec<StorageEncodedBox<Vec<bool>> := [Vec::from([true, false, true]), Vec::default(), Vec::from([true])],
```
As `Vec` is a heap type and is supported, wouldn't this obsolete `StorageVec`? In what case would I use one over the other? How would we ensure that developers know which to use?

I am very much so liking the use of `write_deep_clear` and it's associated functions as well as `StorageLayout`

IGI-111:
@bitzoic A big part of this change is trying to separate and give a proper interface between the structure of what is being stored and the storage method itself.

This is why we can't stay with implicit `u64` fields: you should tell the compile how you want things stored, and indeed you should actually have to make that decision. Encoding has benefits but it also has a cost you may not want to pay, etc.

Being explicit about the storage types also has the nice benefit that `storage.a` is actually of the type of field `a` in `storage`, which means you can easily mentally model what's happening with no magic.

It also means that we do not need to be opinionated as to how people store things, and they can build their own storage data structures as libraries (as is already being done by some users).

rfcs/0013-configurable-and-composable-storage.md
ironcev: Removing the enum and the `Null` and using `Option` instead I like a lot!

Actually, the first version of the `StorageRef` was a struct type-safely wrapping the `StorageKey` but then I ended up thinking how to support not having a reference and introduced `Null` motivated by the similar programming language concept. The result was putting the burden of checking for `Null` when dereferencing _always_, even in the situations where the reference will always be there. Using `Option<StorageRef>` where the `StorageRef` is a struct is definitely the right way!

To the question of having the dedicated `StorageRef`, or using the general `StorageKey` directly, the benefit of having the `struct StorageRef<TStorage>` is there, because of these two reasons:
1. Type-safety.
1. Having easy-to-use, zero-(additional)-cost abstraction that clearly communicates the intent and provides a standardized usage pattern.

Ad 1) `StorageRef<TStorage>` referrs to a storage of a particular type and that is type-checked by the compiler. It also clearly communicates that it referrs to a storage of a particular type. Using `StorageKey` would be more like having a raw pointer into the storage.

Ad 2) Using the naked `StorageKey` will require from developers to implement their own `deref` equivalents every time they use it for referencing. Those will be cumbersome and error-prone, because they will need to use the proper way to create the proper storage type etc. In the worse case, we will end up having that code in contracts, and in the best case, having it packed in various libraries. Having the `StorageRef` provided by the standard library gives a simple type that clearly communicates the intent and gives all convenience in a standardized and zero-cost way.

rfcs/0013-configurable-and-composable-storage.md
ironcev: Thanks for the proposal! I am not happy with the term `Config` either. While the `Value` as the name of the associated type is intuitive, `Config` does not carry such clarity.

Here we need to be cautios of the overall "initialization" vs "configuration" confusion. Since `configurable`s and `storage` are actually _configured_ in code and that configuration can be changed at deployment time, the term "configuration" is consistently used everywhere in the RFC. Sticking to that term will help to remove existing confusion, where the choice of the syntax and wording implies "initialization" and "assignment" to something that is rather a "configuration".

E.g., the `Storage::init` constructor on the other side, is an "initialization" because it will create a storge type instance at the runtime and actually stores the provided values into the storage.

To the origin of the current, non-optimal name `Config`, which I tried to change (my thinking below). Since both the associated type `Config` and the `StorageConfig` are used only by the compiler during the configuration, I've kept the name `Config` although I don't like it.

Alternative names I've played with for the `Config` associated type were:
- `ValueLayout`: because it provides the information on how the value (described by the `Value` type) is layed out in the storage. The reason I didn't choose it is, that storage types can store additional information as well, e.g., `StorageVec` its length, and not only what is in the `Value`.
- `StorageLayout`: it clashed with the `StorageLayout` enum.
- `StorageContent`: because it describes which content is stored in the storage and where exactly.

Using "Init" in the name, as explained about, can lead back to the current confusion.

Perhaps `SlotsConfig` because it provides the information on how to configure the slots, or even better `SlotsContent`?

Additional proposals are welcome :smile:

ironcev: Also, regardless of the terminology we chose at the end, I like the proposed formulation (but to change "initializing" to "configuring"):

> StorageConfig represents a single address initializer for consecutive slots and that Config is a whole, potentially composed, initializer

Hmmm, perhaps:

`Storage(Address)Config/Content` represents a configuration for storing a single `value/content` in consecutive slots starting at the address specified by the `storage_key`. The `SlotsConfig/Content` is a whole, potentially composed, configuration, for storing the entire value contained in the `Storage`.

Whatever the final terminology will be, it should sound well in that sentence :smile:

rfcs/0013-configurable-and-composable-storage.md
ironcev: Agree!

Encouraging and/or discouraging usage is a separate topic, similar to the one of using raw pointers in contract code. It is to be reflected in the way how we convey and document the library/language features, patterns, etc. But in both cases, even discouraging the usage does not justify the name "internal".

The current module name is `storage_api` which we can keep. But, I am more for `raw_api` or some other classifier which will better convay that this is the low-level API, rarely needed, if at all, in the contract code.

@bitzoic Do you have a view or proposal here?

rfcs/0013-configurable-and-composable-storage.md
ironcev: I say - no :smile: (to the negative implications, not to the discussion :smile:)

Or at least I don't see any negative impact on the performance, compared to the approach we have now. On the contrary:
- Performance was in focus and considered carefully. It resulted, e.g, in the `StorageLayout` enum being introduced.
- Just using the `StorageLayout` information consistently will help `Storage` implementors to provide implementations that optimally pack and store data across the whole hierarchy of the stored data.
- Having the convention for "special behavior" methods should encourage providing optimized versions of methods where high-performance is needed and safety guaranteed by surrounding code.

But it can be that I am missing something, so yes we can surely discuss it further. And of course reviewing the proposal from the performance standpoint is highly encouraged.

rfcs/0013-configurable-and-composable-storage.md
ironcev: I like the verb `make`! I didn't like `get` and was thinking about alternatives, but didn't like them, e.g. `create`, either. They all sounded very "runtimeysh". `make` somehow does not have that apeal. Here is your self key and the value to store, _make_ me the <config (good name still to find :smile:)>.

As mentioned, `initializer` can mislead and in the and, the final name behind `make` will depend on how we name the `Config` associated type.

rfcs/0013-configurable-and-composable-storage.md
ironcev: Well, `self_key` _is_ the best name :smile: IMO, of course.

It very precisely communicates what it is, and that's why it is introduced and consistently used in the RFC:

> An important term to establish will be the _self key_. Self key is the `StorageKey` at which an instance of a storage type is stored.

Why I see it having properties of a perfectly good name:
- In the implementations of `Storage` types, the term `self_key` immediately communicates that "this is _my_ storage key, the storage key at which _I_ am stored".
- The public `self_key()` method is equally clear. It gives a clear answer to the question "what is _your_ key?". (We need this method for, e.g., referencing or any other scenario where the knowledge of the slot location of the storage type instance is needed.)
- In the documentation and communication, being reflexive it immediately points to the storage instance we are talking about. 

Any other word that does not imply reflexive relationship (like "self", or "me") will be to broad and possibly applicable to several instances and the question "whose, of what" will not be immediately incorporated in the word.

E.g., having the `field_key` everywhere rises the question "which field? whose field?". The sentence "Field key is the `StorageKey` at which the instance of a storage type is stored" calls for explaining why the word "field". We have to keep in mind that storage type instances can be used independent of the `storage` which makes the term "field" even less understandable and hard to justify.

Also, some storage types could have clashing terminology with the word we choose. E.g., if we chose "element key", it will clash with what an element in the `StorageVec` is.

That's why I am for the name that implies reflexive relationship and see "self" as the perfect candidate. It also corresponds to what "self" is in Sway types. Something that points to, or denotes, myself.

rfcs/0013-configurable-and-composable-storage.md
ironcev: # TL; DR

The proposal (for the next discussion round :smile:) is:
- `fn <operation>() -> Option<T>`: `None` in case of uninitialized storage or other, semantic, error.
- `fn <operation>_or_revert() -> T`
- `fn <operation>_<special_behavior>() -> Option<T>`: `None` in case of uninitialized storage or other, semantic, error.
- `fn <operation>_<special_behavior>_or_revert() -> T`
- `fn <operation>() -> T`: operation never fails.
- `fn <operation>_<special_behavior>() -> T`: operation with special behavior never fails.

Looking at the storage types we have, there will never be an explosion of those variants per operation, and we still have a clear guidelines that capture the overall complexity (explained below).

# Detailed Analysis

After spending a good portion of the last week walking in circles and pondering how to capture the complexity of the storage access semantics into consistent API guidelines, here I am again, doing the same :smile:

I am completely fine with promoting the `Option` returning operations to be the default ones.

But we still need to capture the overall matrix of possible combinations. Since [FuelLabs/sway/#4740](https://github.com/FuelLabs/sway/issues/4740) is not going into details there, let's do it here.

Before starting, just to emphasise the difference, the term "unchecked" in the #4740 and in the RFC paragraph, means two different things. In the #4740 it is a synonim for unwrap while in the RFC it represents a "special behavior". In the concrete example, running the `StorageVec::get()` without checking the vector length and out-of-bounds access.

To the matrix of possible combinations and what we want to capture in the API.

Every operation can potentially:
- fail if the storage is not initialized. E.g., `StorageBox::read()` on uninitialized `StorageBox`.
- fail because of a semantically wrong usage. E.g., out of bounds `StorageVec::get()` on initialized `StorageVec`.
- have one or more "special behaviors" that affect what the operation do. E.g., `StorageVec::get()` without lenght checks. E.g., `StorageVec::pop()` with deep clear.
  - we expect a small number of special behaviors per operation, very often none or mostly one.
  - special behaviors might require additional trait constraints. E.g. `DeepClearStorage`.
  - special behavior operations can also fail because of the first two points.
- have none of the above; neither it can fail nor it has any special behaviors. E.g., `StorageMap::get()`.

The guidelines should reflect the above matrix.

What RFC proposes, to simplify the matrix and the API, and assuming that we are not losing any valuable information there, is to treat the access to the uninitialized storage as a semantically wrong usage and not as a special technical error.

Means, every storage type assigns a meaningful default semantics to the uninitialized state. E.g, an uninitialized `StorageVec` behaves the same as an empty `StorageVec`. For all the storage types we currently support, this is possible in an intuitive way.

It also has a benefit of just creating an instance with `new()` without the need for initialization. `StorageVec::new()` behaves the same as `StorageVec::init([])` but does not produce any storage writes until used. For storage types I've played with and the sample `StoragePair` having this meaningful default behavior was always possible.

If we do not join those two points and consistently want to distinguish between the failing because of the storage being uninitialized and because of a semantically wrong usage, we will necessarily end up in the cumbersome API like, e.g., `StorageVec::get()` returning `Option<Option<T>>` where the first option means failure on uninitialized and the second out-of-bounds. I doubt that developers will ever want to distinguish those two cases.

One more thing to consider in the guidelines is:
> it should be clear when an operation potentially reverts

How to achieve this via naming convention only? (We are talking here about the computational effects. We had this discussion internally triggered by the `#[storage]` attributes not being a part of the function type.)

One possibility would be:
- `<operation>() -> Option<T>`: does not revert.
- `<operation>_<unchecked or some better name>() -> T`: potentially reverts.
- `<operation>() -> T`: does not revert. E.g., `StorageMap::get() -> T`.

And all this should also be aplicable to "special behavior" variants.

Using the term "unchecked" as a suffix I see as misleading. Because, in the case of, e.g., `StorageBox::read_unchecked()` it actually _does check_ and it _will revert if that check fails_.

"unchecked" as a word implies that we are actually skipping checks and who knows what can happen. Developers responsibility. E.g., in the RFC, the `StorageVec::get_unchecked()` cannot fail because it literally _does not check_ the bounds and returns the calculated element. Accessing that element later on can fail, but the `get_unchecked()` itself cannot.

I am arguing here to adopt some other terminology for the `<operation>_<give me the result or revert if it is not there>()`, and not not to have that variant.

Perhaps `<operation>_or_revert()`?

To wrap it up, the proposal (for the next discussion round :smile:) is:
- `fn <operation>() -> Option<T>`: `None` in case of uninitialized storage or other, semantic, error.
- `fn <operation>_or_revert() -> T`
- `fn <operation>_<special_behavior>() -> Option<T>`: `None` in case of uninitialized storage or other, semantic, error.
- `fn <operation>_<special_behavior>_or_revert() -> T`
- `fn <operation>() -> T`: operation never fails.
- `fn <operation>_<special_behavior>() -> T`: operation with special behavior never fails.

Looking at the storage types we have, there will never be an explosion of those variants per operation, and we still have a clear guidelines that capture the overall complexity.

ironcev: A question that immediately comes to my mind is why `_or_revert()/unchecked()` at all, when we can simply call `unwrap()`?

The `<operation>() <-> try_<operation>()` pairs are there from the beginning if that's a valid reason. `<operation>().unwrap()` results in a potentially much bigger code then `<operation>_or_revert()` (see [FuelLabs/sway/#5982](https://github.com/FuelLabs/sway/issues/5982)), but this should be handled in optimizations.

Also, other APIs, aside from the storage API, does not have such pairs. E.g. `Vec` has only `get() -> Option<T>`.

So why to have `_or_revert()/unchecked()` so prominently in the storage API?

SwayStar123:
@ironcev Is there any situation in which i can use StorageBox and would want to use StorageEncodedBox instead?

ironcev:
> Is there any situation in which i can use StorageBox and would want to use StorageEncodedBox instead?

@SwayStar123  I am not aware of any. In the proposal, [this line](https://github.com/FuelLabs/sway-rfcs/blob/40c4dba32b68e55f34d16ea98b782953885371b2/files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw#L31):
```Sway
impl<T> !Storage for StorageEncodedBox<T> where T: Serializable { }
```
forbids using `StorageEncodedBox` if the type can be stored using `StorageBox`.

However, I've left [the discussion question](https://github.com/FuelLabs/sway-rfcs/blob/40c4dba32b68e55f34d16ea98b782953885371b2/files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw#L25) asking the same as you, is there a valid reason to allow that and treat it as a warning rather then forbidding it:
```Sway
// TODO-DISCUSSION: Shell we forbid encoded-boxing serializable types and thus force
//                  them to be boxed in `StorageBox` or should this only be a compiler warning?
//                  Essentially, if a type is `Serializable` encoding it unnecessarily
//                  is a huge waste of computational and storage resources.
```
Note that we can always start as proposed, having a compiler error if a serializable type and thus `StorageBox`able type is attempted to be stored in `EncodedStorageBox` and later on remove the restriction and replace it with a warning, without having a breaking change.

ironcev:
> The RFC also mentions that StorageBox is more gas efficient than StorageEncodedBox, the question I have Is why is it more efficient? How would Serializable trait work? We cannot copy the contiguous memory if we assume the memory representation of types can change in the future.

@esdrubal Currently, this is exactly what we do when storing types into storage, copying memory representation as is. At compile time storage configuration uses the Sway memory layout information to calculate the slots. The `Serializable` marker trait puts the safety around copying memory, ensuring that we do not copy types that, e.g., contain pointers. The similar guard is there when we autoimpl `AbiEncode`.

If the memory representation changes in the future, the issue might indeed arise if we e.g. have code accessing storage that gets LDCed into a contract that used old memory layout. This is a valid concern. I see it being in the same category like some others where we traded undefined behavior for performance. Paying for decoding/encoding at every storage access would definitely be a very high cost to pay for a change of memory layout that we will potentially never introduce. Also, note that we would have a similar issue with the `StorageEncodedBox` if the encoding changes from `v1` to `vNext`.

That's why I am for addressing this concern as a breaking change in existing contracts if it ever happens.

As a side note, what we might introduce in the future is the configurable memory layout equivalent to [#[repr(...)] in Rust](https://doc.rust-lang.org/nomicon/other-reprs.html): [FuelLabs/sway/#5286](https://github.com/FuelLabs/sway/issues/5286)

In that case, we would still be copying the memory. The compiler would be aware of the `#[repr]` attribute and during configuration calculate the slots accordingly. The above concern would remain or be even more prominent. E.g. having the same `struct` but with two different `#[repr]`s in different contracts accessing the same storage.

CC: @IGI-111

SwayStar123:
> > Is there any situation in which i can use StorageBox and would want to use StorageEncodedBox instead?
> 
> @SwayStar123 I am not aware of any. In the proposal, [this line](https://github.com/FuelLabs/sway-rfcs/blob/40c4dba32b68e55f34d16ea98b782953885371b2/files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw#L31):
> 
> ```sway
> impl<T> !Storage for StorageEncodedBox<T> where T: Serializable { }
> ```
> 
> forbids using `StorageEncodedBox` if the type can be stored using `StorageBox`.
> 
> However, I've left [the discussion question](https://github.com/FuelLabs/sway-rfcs/blob/40c4dba32b68e55f34d16ea98b782953885371b2/files/0013-configurable-and-composable-storage/sway-libs/storage/storage_encoded_box.sw#L25) asking the same as you, is there a valid reason to allow that and treat it as a warning rather then forbidding it:
> 
> ```sway
> // TODO-DISCUSSION: Shell we forbid encoded-boxing serializable types and thus force
> //                  them to be boxed in `StorageBox` or should this only be a compiler warning?
> //                  Essentially, if a type is `Serializable` encoding it unnecessarily
> //                  is a huge waste of computational and storage resources.
> ```
> 
> Note that we can always start as proposed, having a compiler error if a serializable type and thus `StorageBox`able type is attempted to be stored in `EncodedStorageBox` and later on remove the restriction and replace it with a warning, without having a breaking change.

Im confused then,  the stated reason for having explicit storagebox and storageencodedbox is that the developer should make a conscious decision to choose between them, but apparently theres no choice at all?

ironcev:
Here are some additional remarks and concerns coming from discussion with @SwayStar123.

1. _StorageBox and EncodedStorageBox. Unnecessary for the developer to have to think about this. Can't the compiler just figure out which one is necessary?_

    The reasoning for explicitly using `StorageBox` and `EncodedStorageBox` are already discussed in detail here:
    https://github.com/FuelLabs/sway-rfcs/pull/40#issuecomment-2159814654
    https://github.com/FuelLabs/sway-rfcs/pull/40#issuecomment-2162942775

    I would like to additionally point out, that in general, developers will actually not need to think about the existance of two stroage boxes. In vast majority of cases, it is simply going to be the `StroageBox`. We will need to properly convey the mental model of our storage approach. Storage types offer structural view over the storage creating hierarchies ending up in atomic storage boxes. Those will, for the majority of cases be the `StorageBox`.

    If the `EncodedStorageBox` is needed, and a developer still habitually picks `StorageBox`, the compiler error will clearly explain why the encoded version is needed. Expressive diagnostics enable us to help developers learn on their own code.

    My point is, in general, there will be no real need to think much upfront about the choice between two storage boxes.

1. _The fact that there are non interchangable vec and storagevec. Strongly against this design. Not only will this be hard to communicate to all devs, it's also adding complexity as compared to solidity, is it not supposed to be the opposite?_

    This is a direct consequence of two design decisions in Sway as a language and our approach to storage, established with Storage v2.0.
    - First, unlike dynamic arrays in Solidity, Sway's vectors are not a language construct but rather a library type. From the compilers perspective, there is no any difference between the `Vec` and any other ordinary Sway struct. `Vec` is just a struct that internally does memory allocation.
    - Second, Sway transparently presents the nature of the contract storage. It has the explicit `storage` declaration to declare storage "variables" and insist on reading and writing to storage via methods that clearly convey that they might fail. The storage access is intentionally conceptually presented as IO access unlike accessing variables in memory.  

    The combination of these two design choices makes `StorageVec` and `Vec` different in the API and not interchangeable. Even if we would built-in special support for `Vec` into compiler (not something I would advocate) and compile it differently when in memory and in storage, `Vec`'s API as such wouldn't fit to storage API.

    The perceived complexity, which I fully agree is there, needs to be communicated through the lenses of the benefits of the Sway approach to storage. Here I like how @IGI-111 put it, "no magic" + "flexibility":
    > Being explicit about the storage types also has the nice benefit that storage.a is actually of the type of field a in storage, which means you can easily mentally model what's happening with no magic.
    > It also means that we do not need to be opinionated as to how people store things, and they can build their own storage data structures as libraries

1. _I can imagine many devs not even knowing storagevec existing and using horribly inefficient code where they read and write the entire vecs everytime they want to update an element_

    With a strong background in a managed language simple to use in an unperformant way, I also see this as a problem.

    The RFC pays a lot of attention to providing the infrastructure and the API that will be difficult to use in a false way, both when defining stroage types and using them. This is one example where, purely throught the API it is not possible to guide the developers to the right choice. On the other side, treating `Vec` differently on the compiler side, or forbidding storing dynamic structs is also not an option.

    The question is what to do to mitigate the potential unintended misuse.

    I see two complementary approaches here. The documentation will have `StorageVec` as the default narrative, as it is now. Also the `StorageBox`. The `EncodedStorageBox<Vec>` should be on the margins and explained with red flag and as potential optimization and very likely not the regular case. In addition, we've discussed standardized static analysis tool similar to Clippy for Sway. E.g., [Sway Analyzer](https://github.com/ourovoros-io/sway-analyzer) can give a warning when using encoded `Vec` in storage.

    But to come back to the flexibility point, storing `EncodedStorageBox<Vec>` can be the more performant option in certain scenarios. E.g., small vectors of `u8` that need to be loaded and stored completely, etc.

ironcev:
> Im confused then, the stated reason for having explicit storagebox and storageencodedbox is that the developer should make a conscious decision to choose between them, but apparently theres no choice at all?

Fair point :smile: We've discussed the topic of choice and the need for both the `StorageBox` and the `StorageEncodedBox` in the team meeting. The discussion addresses this question and solves as well the open TODO-DISCUSSION point that was asking if we should forbid storing serializable types in the `StorageEncodedBox`.

Please see the [Explain difference between `StorageBox` and the `StorageEncodedBox` commit](https://github.com/FuelLabs/sway-rfcs/pull/40/commits/f23d9838a6d3e4896ddaf14aa3990960fde3043a) for the conclusion.

CC: @SwayStar123, @esdrubal, @IGI-111

ironcev:
Hmm, the two dead links in the `collection_context` RFC actually exist, both the old 2019 ones, and the newer, but the link verification still fails. 

```
[✖] https://papl.cs.brown.edu/2020/types.html#(part._.Recursive_.Datatype_.Definitions) → Status: 0
[✖] https://papl.cs.brown.edu/2020/Type_Inference.html → Status: 0
```

Nazeeh21:
By extending the convo with @ironcev from Slack I would like to propose a concept/implementation for the StorageMap concept. I have put all the info regarding that in the gist [here](https://gist.github.com/jecikpo/f4508ab48ef91ca42866fcc1645998a4#file-harness-rs) along with the example implementation. It might be a bit clumsy right now, but at least it presents the idea. I reworked a bit the original `StorageKey` concept and combined it with the existing `StorageMap`, the outcome is a single struct. 

I think it should also support nested maps, but need to test it. 

The API of this `SecureStorageMap` thing is a bit tricky to use as methods and constructors return `Result`, so maybe that could be somehow improved. Let me know what you think about it.

PR #41, opened 2024-08-05
formalize how to introduce changes into the compiler

xunilrj:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/xunilrj/lifecycle-changes/rfcs/0013-changes-lifecyle.md)


rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+
+The compiler needs a guide on how to introduce changes to improve developer experience every time a compiler is updated.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## PR changes
+
+Every PR will have a section named "Release Notes", which will contains a user friendly explanation of what the PR contains.
```
IGI-111, line 24: We should probably introduce the rest of the changeset benefits, including specifying the level of change the PR does (minor, major, patch, etc).

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+
+The compiler release notes **WILL** be the amalgamation of these section "as is".
+
```
IGI-111, line 27: I also think that we should discuss formalizing what our policy towards breaking changes is here.

Some of this is getting discussed elsewhere for other moving parts of the ecosystem, but I think we should commit to not making breaking changes outside of specific timed releases and emergencies (security related bugs, etc).

The normal process should be that a breaking changes goes through behind a feature flag and those features get stabilized at predictable intervals when we do a breaking release.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+This RFC intends to propose a scheme to track changes in the compiler. These changes will be categorized into two ways:
+
+1 - Non-breaking changes;
+2 - Breaking changes.
+
+Breaking changes will be any changes that, after updating the compiler version, causes:
+
+1 - Code without deprecated warning to be flagged as error;
+2 - ABI json properties to be removed or have their type changed;
+3 - SDK (rust, typescript and others) to break;
```
IGI-111, line 73: This includes it implicitly but we should explicitly mention noncompatible changes to the ABI (not the spec, the actual binary interface of the contracts) and VM incompatibility.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+4 - Receipt parsers to break;
+
+The following changes are **not** considered to be breaking changes:
+
+1 - Code flagged with deprecated warning to be flagged as error;
+2 - Code that required an experimental flag, but now are the default behavior to compile;
+3 - Contract ID to change;
+4 - new ABI json properties;
+
```
IGI-111, line 82: This should also include IR changes. (at least for now)

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+1 - Code without deprecated warning to be flagged as error;
+2 - ABI json properties to be removed or have their type changed;
+3 - SDK (rust, typescript and others) to break;
+4 - Receipt parsers to break;
+
+The following changes are **not** considered to be breaking changes:
+
+1 - Code flagged with deprecated warning to be flagged as error;
+2 - Code that required an experimental flag, but now are the default behavior to compile;
```
IGI-111, line 79: This I don't agree with, stabilizing an experimental flag should be a breaking change, if it is indeed noncompatible.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+3 - Contract ID to change;
+4 - new ABI json properties;
+
+## Minor Version Updates
+
+When a developer updates to a different minor version, it is expected no breaking changes, except for severe bugs.
```
IGI-111, line 85:
```suggestion
When a developer updates to a different minor version, no breaking changes are expected, except for severe bugs.
```

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+
+Minor changes will consider these to be breaking and non-breaking changes:
+
+### Breaking Changes 
+
+1 - If any new warning, error or other diagnostic are emitted;
```
IGI-111, line 91: This I don't agree with. Adding new warnings and errors should not necessarily be a breaking changes.

New warnings should always be a nonbreaking change.

New errors should only be a breaking change if previously valid code is now considered erroneous. Invalid sway code that produces new or different errors is not a breaking change.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+2 - If any ABI json properties are removed or have their type changed;
+3 - If any code using SDK (rust, typescript and others) stop working;
+4 - If any receipt cannot be parsed anymore;
+5 - If any "contract id" changes;
```
IGI-111, line 95: This we should discuss. I don't think a contract compiling to a new contract id should necessarily be considered a breaking change. There is an argument for it but I don't see it as breaking because the code still works and compiles as intended.

If you want reproducible builds, you should be using the same exact version of the compiler anyways. So I don't see too much of a benefit in making this breaking.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+6 - Semantic changes
```
IGI-111, line 96: This is too vague. Should be explicited as producing different runtime behavior for the same code. Excluding bugs.

rfcs/0000-changes-lifecyle.md
sdankel: Have we considered putting the release notes into a changelog file, rather than having them spread across different PR descriptions?

rfcs/0000-changes-lifecyle.md
sdankel: New errors could require sway devs to "migrate" to the new release of the compiler, which I think is the definition of a breaking change. New warnings are not necessarily breaking, but for projects using `--error-on-warnings`, they are essentially the same as new errors.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+1 - Code without deprecated warning to be flagged as error;
+2 - ABI json properties to be removed or have their type changed;
+3 - SDK (rust, typescript and others) to break;
+4 - Receipt parsers to break;
+
+The following changes are **not** considered to be breaking changes:
+
+1 - Code flagged with deprecated warning to be flagged as error;
+2 - Code that required an experimental flag, but now are the default behavior to compile;
+3 - Contract ID to change;
```
sdankel, line 80: This is also listed under Breaking Changes 😆

rfcs/0000-changes-lifecyle.md
sdankel: I agree with this, but am curious what the SwayEx devs would say. It would break some of our tests, which means it possibly breaks other sway dev's integration tests.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+4 - new ABI json properties;
+
+## Minor Version Updates
+
+When a developer updates to a different minor version, it is expected no breaking changes, except for severe bugs.
+
+Minor changes will consider these to be breaking and non-breaking changes:
+
+### Breaking Changes 
+
```
sdankel, line 90: It might be good to start off this section with a definition of breaking changes, before listing the categories of breaking changes. Agreeing on a definition will make it easier to debate the categorization.

Here's a definition to start with.

```suggestion
### Breaking Changes 

A breaking change is a difference in functionality from the previous version of the compiler that may require an update to sway code in order for it to compile.
```

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+2 - If any ABI json properties are removed or have their type changed;
+3 - If any code using SDK (rust, typescript and others) stop working;
+4 - If any receipt cannot be parsed anymore;
+5 - If any "contract id" changes;
+6 - Semantic changes
+
+### Non-breaking Changes
+
+1 - new ABI json properties;
+2 - Changing one warning by another, or one error by another;
```
sdankel, line 101: I think this is clearer:

```suggestion
2 - Changing the name or contents of an error or warning;
```

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,205 @@
...
+3 - Code that was **NOT** compiling, but now is.
+
+## Major Version Updates
+
+Major update is the opportunity to introduce breaking changes, but they cannot be introduced abruptly, and depending on the change, will need to be introduced into phases.
+
+Changes that only affect the compiler/language, can be introduced freely following sway pace. Changes that interact with others tools, will need to be introduced into phases and in coordination with others tools.
+
+These changes will need to be behind "feature flags", which will be described later.
```
sdankel, line 110:
```suggestion
These changes will need to be behind "feature flags", which are described above.
```

rfcs/0000-changes-lifecyle.md
xunilrj: My though is aligned with what @sdankel said. 

I expect every team to treat warnings as errors, and in these cases, new warnings will force developers to change code.

For the moment I moved this to the "open questions" section.

rfcs/0000-changes-lifecyle.md
xunilrj: For the moment I moved this to the "open questions" section.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,184 @@
...
+- How scripts/predicates return data from `main`;
+- How the `contract method selector` is encoded;
+- How `contract method` arguments are encoded;
+- How `log` data is encoded;
+- How `message` data is encoded;
+
+  4 - Utilization of new VM opcodes;
+  5 - IR changes
+  6 - Receipt parsers to break;
+  7 - When a compiler feature or a standard library produce different behavior for the same code (semantic changes);
```
Voxelot, line 60:
```suggestion
- Utilization of new VM opcodes;
- IR changes
- Receipt parsers to break;
- When a compiler feature or a standard library produce different behavior for the same code (semantic changes);
```

formatting nits

rfcs/0000-changes-lifecyle.md
sdankel: Another option is to generate the changelog file from commits as part of the release process, using something like https://git-cliff.org/

We'd have to be more rigorous about commit messages, i.e. using `feat!:` to indicate a breaking change for a feature.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,184 @@
...
+## Breaking changes
+
+A breaking change is a difference in functionality from the previous version of the compiler that may require an update to sway code in order for it to compile.
+
+The following changes are defined to be breaking changes, and will need to follow the process of being gated by features flags.
+
+1 - Code without deprecated warning to be flagged as error;
+2 - ABI json properties to be removed or have their type changed;
+3 - Binary encoding that will break how SDK (and others) communicate with the compiler, that includes
+
```
Voxelot, line 49:
```suggestion
- Code without deprecated warning to be flagged as error;
- ABI json properties to be removed or have their type changed;
- Binary encoding that will break how SDK (and others) communicate with the compiler, that includes
```

formatting nits

rfcs/0000-changes-lifecyle.md
Voxelot, line 60: The formatting seems really off here:

```suggestion
1. Code without deprecated warning to be flagged as error;
1. ABI json properties to be removed or have their type changed;
1. Binary encoding that will break how SDK (and others) communicate with the compiler, that includes
   - How scripts/predicates accept arguments on `main`;
   - How scripts/predicates return data from `main`;
   - How the `contract method selector` is encoded;
   - How `contract method` arguments are encoded;
   - How `log` data is encoded;
   - How `message` data is encoded;
1. Utilization of new VM opcodes;
1. IR changes
1. Receipt parsers to break;
1. When a compiler feature or a standard library produce different behavior for the same code (semantic changes);
```

xunilrj: Done.

ironcev:
Proposal is to add additional sub-chapter or a paragraph under the _Feature flags_ chapter. E.g., _Feature flags and `#[cfg]` attribute_.

Every experimental flag provides an argument that can be used inside of the `#[cfg]` attribute. This enables conditional compilation of Sway code only if the experimental feature is opted in. The argument is formed out of the feature flag name prefixed with the _experimental_. E.g., `#[cfg(experimental_composable_storage)`.

We need this support, because developing a new feature does not mean only extending the compiler, but often also the standard libraries. E.g., when developing new storage, we will have different implementations of storage types like, e.g., `StorageVec`. We want those implementations to coexist with those expected by the current implementation of storage in the compiler. Supporting `#[cfg]`, of course, goes beyond Sway standard libraries. 3rd party libraries can be adapted as well to continuously follow a new feature development. Also, we had the situation already that tests needed to be conditionally compiled to support different contract IDs for encoding v0 and v1.

ironcev:
Proposal is to explicitly mention that we will not support Sway Editions, the way Rust does it. Once an experimental feature gets promoted to standard behavior, it cannot be turned off anymore in that version and newer versions of the compiler. In other words, it is not possible to take a compiler of a certain version and "ask" it to behave the same as some older version.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,220 @@
...
+## How `rustc` does it
+
+https://rustc-dev-guide.rust-lang.org/implementing_new_features.html#stability-in-code
+
+1. - Open a tracking issue;
+2. - Pick a name for the feature gate;
+3. - Add the feature gate to the compiler code;
+4. - If the feature gate is not set, you should either maintain the pre-feature behavior or raise an error, depending on what makes sense;
+5. - Add a test to ensure the feature cannot be used without a feature gate;
+6. - Add a section to the unstable book;
```
ironcev, line 190: Do we also want to have _The Sway Unstable Book_? IMO, yes. It will be easier to communicate to devs what are we working on, but even more important, for us, we can craft the feature documentation as we go and save us from stress of documenting everything on the end of the development.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,256 @@
...
+```
+ReleaseNotes.md merge=union
+```
+
+
+## Breaking changes
+
+A breaking change is a difference in functionality from the previous version of the compiler that may require an update to sway code in order for it to compile.
+
+The following changes are defined to be breaking changes, and will need to follow the process of being gated by features flags.
```
ironcev, line 80: Perhaps additional one: changing the way how `storage` field slots are calculated? Right now, we calculate them as `sha256("storage::ns1::ns2::..::nsN.field_name")` and want to introduce a new algorithm: `sha256((0u8, "storage::ns1::ns2::..::nsN.field_name"))`. There were questions on how slots/storage keys are calculated for `storage` fields, so I assume there are devs who rely on that.

ironcev:
How are we going to approach testing the features on CI?

In a near future we will have several features being developed in parallel, all behind feature flags, all with their code in master, either in development or fully developed and waiting for becoming the default behavior. All of them will have their tests or some tests adapted to work with and without the feature.

Can we elaborate more on this in the RFC? Some of the topics:
- how features will be tested in CI?
- how artefacts that can differ in features like ABI JSON or storage slots will be defined?
- how will different features be conditionally compiled in test code? We will need both #[cfg] and #![cfg].

rfcs/0000-changes-lifecyle.md
xunilrj: Moving this to the open questions section.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+LSP can take advantage of specific error messages, and suggest user to enable the corresponding feature.
+
+## CST, AST and Typed Tree
+
+All trees must always support new features, which means that new nodes will always exist. Their specific behavior, desugaring, etc... will be gated by the experimental feature.
+
+More specifically, this means that variants should not be behind compiler flags.
+
+## Tests
```
ironcev, line 219: On the CI, are we going to run tests for all the features? What about the combinations of the features? We don't need to define that in this RFC, perhaps it's too early. But it will be an important practical question. One possibility would be to have a CI run with all experimental features off, and one with all on.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+To maximize user experience when updating the compiler, we need a guide on how to introduce breaking changes.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The compiler will follow a "rolling release" scheme, which means that periodically (to be specified) a
+new major version will be released.
+
+This means that as soon as the compiler reaches version "1.0.0", the next **major version** will be "1.1.0"; and
```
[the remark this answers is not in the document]
xunilrj: Hum...  I think `major.minor.patch` makes more sense. Not sure what would be a major increase in our case.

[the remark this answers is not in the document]
xunilrj: I changed to a more neutral wording, just calling for "next version" and "next patch".

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+The only changes that will trigger "minor updates" are:
+1 - Urgent security fixes;
+2 - Bugs that render "stable" functionality unusable;
+
+## Release Notes
+
+To track all "release notes" it will be created a special file `ReleaseNotes.md` that will live in the repo root.
+
+When a version is being launched, for example, `1.10` a new section `1.11-nightly` will be created empty. From this time on, all new PRs will create a new item under the "next version", `1.11` in this example.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+```
+# Version 1.11-nightly
+
+## Forc
+
+### Breaking Changes
+### New Features
+### Bugs Fix
+### Others
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+## Tools
+
+## Sway
+
+## `sway-lib-core` and `sway-lib-std`
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+1 - Urgent security fixes;
+2 - Bugs that render "stable" functionality unusable;
+
+## Release Notes
+
+To track all "release notes" it will be created a special file `ReleaseNotes.md` that will live in the repo root.
+
+When a version is being launched, for example, `1.10` a new section `1.11-nightly` will be created empty. From this time on, all new PRs will create a new item under the "next version", `1.11` in this example.
+
+Each item will follow the template:
```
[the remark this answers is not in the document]
xunilrj: The idea was a user friendly one line description. I have changed the template description below.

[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+[guide-level-explanation]: #guide-level-explanation
+
+The compiler will follow a "rolling release" scheme, which means that periodically (to be specified) a
+new major version will be released.
+
+This means that as soon as the compiler reaches version "1.0.0", the next **major version** will be "1.1.0"; and
+after that and if needed, the next **minor version** would be "1.1.1".
+
+The only changes that will trigger "minor updates" are:
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+1 - Urgent security fixes;
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+that it is not possible to "turn off" a "match expression" or any other language feature.
+
+If the feature contains some configuration or choice, the "default" value will be the default, and other options
+will be available, but it cannot be turned off.
+
+## Conditional compilation
+
+There are cases where conditional compilation will be needed. For these cases, each experimental feature will also have a corresponding `#[cfg(...)]`, like the example below:
+
+````sway
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+#[cfg(experimental_new_encoding = false)]
+const CONTRACT_ID = 0x14ed3cd06c2947248f69d54bfa681fe40d26267be84df7e19e253622b7921bbe;
+#[cfg(experimental_new_encoding = true)]
+const CONTRACT_ID = 0x316c03d37b53eaeffe22c2d2df50d675e2b2ee07bd8b73f852e686129aeba462;
+``
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+# Enabling features on `sway`
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+experimental = ["...", "..."]
+````
+
+or using the CLI
+
+```
+> forc ... --experimental some_feature,another_feature
+```
+
+or event using the environment variables
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+```
+> FORC_EXPERIMENTAL_SOME_FEATURE forc ...
+```
+
+These flags also need to be enabled programmatically by any compiler driver, like tests.
+
+Unlike `Rust` we will not support features inside sway code like the example below, because some features will span across multiple tools. That would demand `forc` to parse, or ask the `sway` compiler if a feature is enabled or not.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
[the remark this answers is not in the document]
xunilrj: `env vars` are normally more useful when one does not have access to the CLI. FOr example: CI, or when one is compiling a project using a SDK that do not allow `forc` to be configured. Then `env vars` become more useful.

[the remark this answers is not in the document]
xunilrj: > Do we fully override or aggregate?

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+}
+```
+
+## Formatting
+
+Formatting should always format new syntax, even when the flag is off. To avoid breaking other parts of the code, or even worse, removing code.
+
+## LSP
+
+LSP can take advantage of specific error messages, and suggest user to enable the corresponding feature.
```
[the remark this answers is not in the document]
xunilrj: Actually went with "suggest user". What do you think?

rfcs/0000-changes-lifecyle.md
xunilrj: What is going to define the configuration is the `test.*.toml`. The default `test.toml` will run the compiler with all experimental off (the same as no flags).

When a specific test needs a specific feature it will have a `test.feature_a.toml`, where that feature will be enabled. Now running our tests will run this specific test twice.

If we force a specific feature we can use the environment variable:

```
> FORC_EXPERIMENTAL=feature_a cargo r -p test --release
```

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+This means that as soon as the compiler reaches version "1.0.0", the next **major version** will be "1.1.0"; and
+after that and if needed, the next **minor version** would be "1.1.1".
+
+The only changes that will trigger "minor updates" are:
+1 - Urgent security fixes;
+2 - Bugs that render "stable" functionality unusable;
+
+## Release Notes
+
+To track all "release notes" it will be created a special file `ReleaseNotes.md` that will live in the repo root.
```
JoshuaBatty, line 33:
```suggestion
To track all release notes, a special file named `ReleaseNotes.md` will be created in the repository's root directory.
```

kayagokalp (commented):
Just left a single question, it is great that this is getting formalized :) 

Out of the scope of this PR but something that we should start thinking about is the is the question that we are putting aside for some time now. The fact that forc and rest of the plugins are getting released at the same time as the compiler does. Which means we do not have any way conveying the breakage in those parts of the released stuff.

To make my point clear, let me give an extreme example:  any change to the dependency declaration syntax etc in the `Forc.toml` would mean all existing projects are no longer building and thus breaking in terms of `forc`, subsequently the end users as well. So marking that `forc` version with a minor bump over the last one (imagine no breaking changes included after last release in terms of the compiler), will convey the idea that the project should still be building. But in reality nothing will build. 

This comment of mine would be completely irrelevant to this PR, if the compiler had any other entry, like `rustc`. As then it becomes the forc's own versioning discussion and totally separate from the compiler. But currently `forc` is the only entrance to the compiler and only way of using it. So it might make sense to discuss this scenario as well for compiler versioning.

rfcs/0000-changes-lifecyle.md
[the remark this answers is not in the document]
alfiedotwtf: This section is possibly the most important part of the document as it deals with how we actually handle the versioning of breaking changes, but I think I'm still a little confused. I'm hoping the following example can help me understand:

Let's say we're currently live at version `X.Y.Z`, and we have a new release ready to go containing the following in its release notes:

```
### Breaking Changes

- This release removes the ABC feature
```

Seeing as the next release will be a breaking change, what will be the SemVer for its release? Will it be:

- (X+1).Y.Z
- X.(Y+1).Z
- X.Y.(Z+1)

In SemVer this release will be `(X+1).Y.Z` but as I read this current section the rolling release will put it at `X.(Y+1).Z`? Whatever the version we use is not my point here - I'm just thinking we need to be more explicit here with whichever way we decide so that it's not ambiguous in the future.

To be more explicit, maybe a new sub-section should be added here called "Semantic Versioning" which summarises how we will categorise/number each type of change i.e something like:

### Semantic Versioning

The compiler will follow a "rolling release" scheme, meaning that each new release (happening periodically but yet to be specified) will steamroll over all previous releases.

#### MAJOR

`MAJOR` number changes are strictly reserved to deal with dramatic differences in the Sway Language.

#### MINOR

Every normal change, which includes **_all_** breaking changes, new features, and bug fixes, will be a bump increase in the `MINOR` number.

#### PATCH

`PATCH` number changes are out-of-band, and are strictly reserved for security fixes to each currently supported `MAJOR.MINOR` release. In other words, if a security fix is needed, every existing supported release will include a new patch bump release.
```

File: rfcs/0000-changes-lifecyle.md [hunk 33]
```diff
@@ -0,0 +1,318 @@
...
+## Should new warnings be considered breaking changes?
+
+1. Normally warnings should not be considered breaking changes, because they do not break anything.
+2. on the other hand, if the team treats warnings as errors, new warnings will break CIs, and demand
+   developers' attention. Not all fixes are trivial and can demand bigger code changes than the user would expect
+   from a minor update from the compiler.
+
+## How to deal with "contract id"?
+
+Should we consider contract ID changes to be breaking changes?
```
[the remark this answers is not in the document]
xunilrj: That is also the general feeling at the moment. 

But I think that there is a chance users will bring two complaints:
1 - update manually is annoying and error-prone;
2 - if the "contract id" changed because of a change I don´t "care/want", I would need to deploy the contract again incurring costs. Or have a complex table like this contract must be compiled with this version, and that contract with that version etc...

We are also considering a tool to help update contract IDs where needed.

rfcs/0000-changes-lifecyle.md
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
```diff
@@ -0,0 +1,318 @@
...
+
+The compiler will follow a "rolling release" scheme, which means that periodically (to be specified) a
+new major version will be released.
+
+This means that as soon as the compiler reaches version "1.0.0", the next **major version** will be "1.1.0"; and
+after that and if needed, the next **minor version** would be "1.1.1".
+
+The only changes that will trigger "minor updates" are:
+1 - Urgent security fixes;
+2 - Bugs that render "stable" functionality unusable;
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0000-changes-lifecyle.md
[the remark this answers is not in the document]
ironcev: I like @alfiedotwtf's detailed wording on major, minor, and patch. Only the title "Semantic Versioning" I find potentially confusing because, at least in my head, it immediately associates on semver where the expectation is to increase the major on breaking changes while in our case, the major version is indeed more a "marketing version". Perhaps just title it "Versioning Scheme"?

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+1. Code without deprecated warning to be flagged as an error;
+2. ABI JSON properties to be removed or have their type changed;
+3. Binary encoding that will break how SDK (and others) communicate with the compiler, including:
+   - How scripts/predicates accept arguments on `main`;
+   - How scripts/predicates return data from `main`;
+   - How the `contract method selector` is encoded;
+   - How `contract method` arguments are encoded;
+   - How `log` data is encoded;
+   - How `message` data is encoded;
+4. Utilization of new VM opcodes;
```
IGI-111, line 98: I think this should be clarified.

Fully additive changes to the opcode set (say by reinterpreting reserved opcodes) should not be considered a breaking change in my opinion.

Only that which changes or breaks existing user code should be breaking in the ideal, new vm additions that purely extend the feature set do not do this.

xunilrj: My idea was similar to IR changes. Tools that read IR/binary opcodes will break with new opcodes. That is why these would need to be introduced as breaking changes.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+[prior-art]: #prior-art
+
+These are the sources and stacks used as motivation for this RFC.
+
+## How `rustc` does it
+
+https://rustc-dev-guide.rust-lang.org/implementing_new_features.html#stability-in-code
+
+1. - Open a tracking issue;
+2. - Pick a name for the feature gate;
```
ironcev, line 308:
```suggestion
2. Pick a name for the feature gate;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+3. - Add the feature gate to the compiler code;
```
ironcev, line 309:
```suggestion
3. Add the feature gate to the compiler code;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+4. - If the feature gate is not set, you should either maintain the pre-feature behavior or raise an error, depending on what makes sense;
```
ironcev, line 310:
```suggestion
4. If the feature gate is not set, you should either maintain the pre-feature behavior or raise an error, depending on what makes sense;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+5. - Add a test to ensure the feature cannot be used without a feature gate;
```
ironcev, line 311:
```suggestion
5. Add a test to ensure the feature cannot be used without a feature gate;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+6. - Add a section to the unstable book;
```
ironcev, line 312:
```suggestion
6. Add a section to the unstable book;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+7. - Write a lot of tests for the new feature;
```
ironcev, line 313:
```suggestion
7. Write a lot of tests for the new feature;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+8. - Get your PR reviewed and land it.
```
ironcev, line 314:
```suggestion
8. Get your PR reviewed and land it.
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+9. - Add a section to the unstable book;
```
ironcev, line 315:
```suggestion
9. Add a section to the unstable book;
```

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+1. In files inside the repo? - Saving in files sounds like the best approach, but if we use the same file, all PRs will conflict. To avoid this we can do like `changeset` and use random names.
+2. In GitHub PR descriptions? - This is the easiest approach, but it can be cumbersome to recover these messages.
+3. In git commit messages? - Given that the commit message is "created" when the PR is merged, there is no way to guarantee that "release notes" will exist or be "parseable" when we need them.
+   - https://git-cliff.org/
+
+## Should new warnings be considered breaking changes?
+
+1. Normally warnings should not be considered breaking changes, because they do not break anything.
+2. on the other hand, if the team treats warnings as errors, new warnings will break CIs, and demand
```
ironcev, line 339:
```suggestion
2. On the other hand, if the team treats warnings as errors, new warnings will break CIs, and demand
```

rfcs/0013-changes-lifecyle.md
IGI-111: Those tools support a particular version of the VM. So long as we produce legal bytecode for that particular VM, I don't think that's a breaking change, even if some are new.

And VM changes are already a breaking change. So yeah I don't think we want it.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+## Standard Libraries
+
+...
+
+# Version 1.10
+
+...
+```
+
+Inside each section, items will follow the template of a user friendly one line description nad link to the PR.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+```
+- Index operator using Index trait [#6356](https://github.com/FuelLabs/sway/pull/6356)
+```
+
+Given that all PRs will touch this file, to avoid conflicts and decrease the experience when merging them, we will use "git custom merge driver" (see https://git-scm.com/docs/gitattributes#_defining_a_custom_merge_driver).
```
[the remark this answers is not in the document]
xunilrj: I will use the link, but I prefer the link text to be visible.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+This allows a custom merge strategy to a specific file using ".gitattributes".
+
+```
+ReleaseNotes.md merge=union
+```
+
+## Breaking changes
+
+A breaking change is a difference in functionality from the previous version of the compiler that may require an update to sway code in order for it to compile.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+8. When `storage` is impacted. Particularly addresses.
+
+## Feature flags
+
+Any complex change that needs to be gated will follow these steps.
+
+1. A specific GitHub issue labeled with `feature-*` and `track-feature`. This issue should be the umbrella for everything related to this feature;
+2. A preliminary PR enabling the feature flag should be created and linked to the umbrella issue; This PR will enable the feature flag `--experimental <comma-separated-list>` on all tools.
+3. As many PRs will be created and merged as normal;
+4. A chapter inside `Sway Unstable book` will be created and updated as needed;
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+If a feature is not turned on by `forc.toml`, it can still be turned on by the CLI and environment variables.
+
+A special token `*` will mean "all features" in the sense that all features can be turned on or turned off.
+
+```
+> forc ... --experimental *
+```
+
+This is specially useful to control what features are enabled
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+# Reference-level explanation
+
+[reference-level-explanation]: #reference-level-explanation
+
+This section contains the suggested guide on how to introduce changes to different parts of the compiler and associated tools.
+
+## `sway-features` crate
+
+A new crate named `sway-features` will be created and will contain **ALL** features and their metadata. The best suggestion is a macro to define features where enums, documentation, etc... will be generated.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+> forc ... --experimental a_feature --a-feature-option 1
+```
+
+## Lexer and Parser
+
+To allow as user-friendly error messages as possible, in all possible cases, we want both the lexer and the parser to parse the new syntax even with the feature off.
+
+After that, the lexer and the parser will mark that an experimental feature was lexed/parsed; and a check will guarantee that no disabled experimental was parsed.
+
+The error message will have a message explaining that the feature is experimental and a Github link for more details on the stabilization lifecycle.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+Formatting should always format new syntax, even when the flag is off. To avoid breaking other parts of the code, or even worse, removing code.
+
+## LSP
+
+LSP can take advantage of specific error messages, and suggest users to enable the corresponding feature.
+
+## CST, AST and Typed Tree
+
+All trees must always support new features, which means that new nodes will always exist. Their specific behavior, desugaring, etc... will be gated by the experimental feature.
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+[prior-art]: #prior-art
+
+These are the sources and stacks used as motivation for this RFC.
+
+## How `rustc` does it
+
+https://rustc-dev-guide.rust-lang.org/implementing_new_features.html#stability-in-code
+
+1. - Open a tracking issue;
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Formatting
+
+Formatting should always format new syntax, even when the flag is off. To avoid breaking other parts of the code, or even worse, removing code.
+
+## LSP
+
+LSP can take advantage of specific error messages, and suggest users to enable the corresponding feature.
+
+## CST, AST and Typed Tree
```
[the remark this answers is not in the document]
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+2. ABI JSON properties to be removed or have their type changed;
+3. Binary encoding that will break how SDK (and others) communicate with the compiler, including:
+   - How scripts/predicates accept arguments on `main`;
+   - How scripts/predicates return data from `main`;
+   - How the `contract method selector` is encoded;
+   - How `contract method` arguments are encoded;
+   - How `log` data is encoded;
+   - How `message` data is encoded;
+4. Utilization of new VM opcodes;
+5. IR changes
```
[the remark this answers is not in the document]
xunilrj: Done.

IGI-111 (approved):

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,349 @@
...
+## Should new warnings be considered breaking changes?
+
+1. Normally warnings should not be considered breaking changes, because they do not break anything.
+2. on the other hand, if the team treats warnings as errors, new warnings will break CIs, and demand
+   developers' attention. Not all fixes are trivial and can demand bigger code changes than the user would expect
+   from a minor update from the compiler.
+
+## How to deal with "contract id"?
+
+Should we consider contract ID changes to be breaking changes?
```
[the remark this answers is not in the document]
IGI-111: I understand the need and I do think it's relevant, but I think that goes beyond what we should do in this RFC.

We can amend it to include something like that later, even switch to contract id changes being breaking, but I don't think it's anywhere near realistic to do so before 1.0.0 when we have so much feature work remaining that's going to break these significantly.

ironcev (approved):

rfcs/0013-changes-lifecyle.md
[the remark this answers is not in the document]
sdankel: Agreed, just wanted to raise this as a potential future use. We might not need it anyway.

sdankel (approved):

JoshuaBatty (approved):

kayagokalp (approved):

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+To maximize user experience when updating the compiler, we need a guide on how to introduce changes.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+The compiler will follow a "rolling release" scheme, which means that periodically (to be specified) a
+new version will be released.
+
+This means that as soon as the compiler reaches version "1.0.0", the next **version** will be "1.1.0"; and
```
Voxelot, line 24: following from the [torvolds line](https://lkml.org/lkml/2012/12/23/75), "WE DO NOT BREAK USERSPACE!" Should we clarify that patches will respect semver expectations? ie. minor releases and patches don't require any user intervention? Unless we are pre 1.0, then minor releases are treated like major breaking ones.

Its not clear to me from the rolling release scheme description here what guarantees - _if any_ - we are providing to our users and how the versions signal changes.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+Any complex change that needs to be gated will follow these steps.
+
+1. A specific GitHub issue labeled with `feature-*` and `track-feature`. This issue should be the umbrella for everything related to this feature;
+2. A preliminary PR enabling the feature flag should be created and linked to the umbrella issue; This PR will enable the feature flag `--experimental <comma-separated-list>` on all tools.
+3. As many PRs will be created and merged as normal;
+4. A chapter inside `Sway Unstable Book` will be created and updated as needed;
+5. When the feature is ready, a closing PR will be created and wait until the feature flag is enabled by default.
+6. On a later date, the feature flag can be removed making the feature the default behavior of the compiler.
+
+Once the feature is merged into `master`, it will not be possible to "turn off" the feature. In the same sense
```
Voxelot, line 114: doesnt need to be specified here necessarily, but if a flag is disabled we should just emit a warning to the user to ensure the presence of flag that is no longer used doesn't break or prevent compilation in any way. If flags are specified that never existed then i think it's fair to throw an error to the user so they know it's not working as they may expect.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+## Feature flags
+
+Any complex change that needs to be gated will follow these steps.
+
+1. A specific GitHub issue labeled with `feature-*` and `track-feature`. This issue should be the umbrella for everything related to this feature;
+2. A preliminary PR enabling the feature flag should be created and linked to the umbrella issue; This PR will enable the feature flag `--experimental <comma-separated-list>` on all tools.
+3. As many PRs will be created and merged as normal;
+4. A chapter inside `Sway Unstable Book` will be created and updated as needed;
+5. When the feature is ready, a closing PR will be created and wait until the feature flag is enabled by default.
+6. On a later date, the feature flag can be removed making the feature the default behavior of the compiler.
```
Voxelot, line 112: .. and the docs should be migrated to the stable portion of the book

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+
+Once the feature is merged into `master`, it will not be possible to "turn off" the feature. In the same sense
+that it is not possible to "turn off" a "match expression" or any other language feature.
+
+If the feature contains some configuration or choice, the "default" value will be the default, and other options
+will be available, but it cannot be turned off.
+
```
Voxelot, line 119: should we also specify anything w.r.t. sway-doc? I.e. do we want to show experimental methods or types and the flags needed to enable them in the auto-generated docs?

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+## Tests
+
+To allow these experimental features to be tested, it is necessary to specify which configurations each test must run.
+Today, by default, a test runs with the "default compiler configuration", but it is necessary to allow tests to
+run with any combination of compiler configuration.
+
+To allow complete tests of any compiler configuration, any configuration that today resides inside `test.toml`
+needs to be configured for each configuration.
+
+An example is the `encoding v1` which has different inputs and outputs from `encoding v0` when testing a script
```
Voxelot, line 254: This section reads more like a possible proposal, whereas RFCs should describe the state of the system as if it's already in place. That way when it is merged and referenced later on, the expectations and specifications are clear to the user.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+that takes arguments.
+
+The easiest way to approach this seems to be to support multiple `test.toml` with a suffix to differentiate them.
```
Voxelot, line 257: This section reads more like a possible proposal, whereas RFCs should describe the state of the system as if it's already in place. That way when it is merged and referenced later on, the expectations and specifications are clear to the user.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+```
+
+The same strategy can be used for snapshot tests. We can use multiple `snapshot.toml` and create different
+snapshots using the file suffix. So `snapshot.feature_a.toml` will generate `snapshot@feature_a.snap`.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+This will increase the complexity of the compiler. Not all flags used to compile end up in the JSON ABI, or other outputs. Which can make reproducibility harder.
```
Voxelot, line 282: this may impact abi verification tools (cc @sdankel)

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+## Should new warnings be considered breaking changes?
+
+1. Normally warnings should not be considered breaking changes, because they do not break anything.
+2. On the other hand, if the team treats warnings as errors, new warnings will break CIs, and demand
+   developers' attention. Not all fixes are trivial and can demand bigger code changes than the user would expect
+   from a minor update from the compiler.
+
+## How to deal with "contract id"?
+
+Should we consider contract ID changes to be breaking changes?
```
Voxelot, line 344: Im more hesitant to say contract ID changes are breaking, as any difference in bytecode output due to some minor optimization would then be considered a breaking change.

rfcs/0013-changes-lifecyle.md
```diff
@@ -0,0 +1,348 @@
...
+[unresolved-questions]: #unresolved-questions
+
+## Where should we save "release notes"?
+
+1. In files inside the repo? - Saving in files sounds like the best approach, but if we use the same file, all PRs will conflict. To avoid this we can do like `changeset` and use random names.
+2. In GitHub PR descriptions? - This is the easiest approach, but it can be cumbersome to recover these messages.
+3. In git commit messages? - Given that the commit message is "created" when the PR is merged, there is no way to guarantee that "release notes" will exist or be "parseable" when we need them.
+   - https://git-cliff.org/
+
+## Should new warnings be considered breaking changes?
```
Voxelot, line 335: I'd say breaking for patches, but not for minor releases. At least this is how rust typically works with clippy warnings etc.

rfcs/0013-changes-lifecyle.md
IGI-111: We should follow the semver guidelines here. Use major for breaking changes, minor for nonbreaking changes and patch for bug fixes only.

A simple mention that we do so and that this document is meant to elucidate what we consider breaking should suffice.

rfcs/0013-changes-lifecyle.md
IGI-111: I think forcing you to update your flags when features are stabilized is a feature, not a bug.

rfcs/0013-changes-lifecyle.md
IGI-111: Surely such tools should only use stable compiler features. We may want to extend support which requires recoding what flags are used on top of the compiler version, but we should not encourage people to use experimental features. In fact this should be more discouraged than it is in Rust given our environment.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: Done. I am now specifying that `forc-doc` will decorate experimental types/fns.

rfcs/0013-changes-lifecyle.md
xunilrj: Done.

rfcs/0013-changes-lifecyle.md
xunilrj: I agree with @IGI-111. This does not mean that we can not generate warnings along the way, before removing the flag.

rfcs/0013-changes-lifecyle.md
xunilrj: I vote for this to be dealt in another RFC.

PR #43, opened 2024-11-25
Abi errors RFC

IGI-111:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/2e0a7a30670e8e404ec5924fed8186e11f487723/rfcs/0014-abi-errors.md)


hal3e (approved):
Great rfc. This should make the devx so much better. 

- I would like to work on this on the rust sdk side.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+        },
+        "logId": "4933727799282657266",
+        "msg": null,
+    }
+    "18446744069414584324": {
+        "pos": {
+          "file": "main.sw",
+          "line": 17,
+          "col": 10
+        },
```
nedsalk, line 272: It would be beneficial to have a `"fn": "function-name"` field in the `pos` object as it'd allow the SDKs to inform the user of the failure modes of the method that's being invoked, be it through some sort of `Result<T, Err>` type or adding comments in the SDK-generated code.

IGI-111: Can you expand on what you mean here? The function the `panic` is within? The contract method that invoked the `panic`?

Given that functions don't exist at runtime and that a given `panic` invocation can be invoked from multiple places, this is ambiguous and may not be decidable information.

I understand the need for some kind of stack trace down the line but I'm not sure how achievable that even is.

nedsalk: > Can you expand on what you mean here? The function the panic is within? The contract method that invoked the panic?

I mean the contract method that invoked the `panic` as that's the thing that users interact with via the the SDK. Perhaps that information could even be put as an array in the function part of the abi:
```
  "functions": [
    {
      "inputs": [
        {
          "name": "a",
          "concreteTypeId": "c89951a24c6ca28c13fd1cfdc646b2b656d69e61a92b91023be7eb58eb914b6b"
        }
      ],
      "name": "main",
      "output": "2e38e77b22c314a449e91fafed92a43826ac6aa403ae6a8acb6cf58239fbaf5d",
      "attributes": null,
      "errorCodes": [ "18446744069414584321", "18446744069414584321", 
                      "18446744069414584322", "18446744069414584323", 
                      "18446744069414584324", "18446744069414584325" ]
    }
  ],
```

nedsalk: > It's not clear to me what an hypothetical fn field would contain here and how helpful it would be to provide that information.

Just the contract function's name so that the SDK can inform the user that the function can panic with the error. This seemed like an opportunity to provide more info to the user, but if it can't be done - it can't be done.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+Moreover, Ethereum style embedding of errors in a `b256` isn't practical for us
+or generally desirable since the FuelVM uses 64 bit error codes and we aim to
+minimize the footprint of errors on deployed contracts.
+
+# Unresolved questions
+
+[unresolved-questions]: #unresolved-questions
+
+It's unclear how much of the existing special error codes produced by the
```
SwayStar123, line 434: I think its fine if we completely remove the existing std lib special error codes.
It will be a breaking change but only really for tests I believe

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+something that can also be used for recoverable errors.
+
+To this end we want to introduce a mechanism that makes error codes an
+understandable part of the ABI, and language features that allow the user to
+define error messages, error types and invoke those errors with the appropriate
+information being generated.
+
+We also want recoverable errors to connect to this mechanism so that all error
+handling in a given Sway program can go through one or a small set of user
+defined types and error strings minimally burden the final bytecode size.
```
maschad, line 37: I'm having trouble understanding this sentence,  does the suggestion below clarify the intention?

```suggestion
defined types and error strings to minimize the final bytecode size.
```

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+- RFC PR: [FuelLabs/sway-rfcs#43](https://github.com/FuelLabs/sway-rfcs/pull/43)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+
+# Summary
+
+[summary]: #summary
+
+This RFC aims to standardize error handling in Sway so that there is a
+deliberate and clear way to produce error messages that are easily consumable by
+language tools and do not burden on chain artifacts with error strings.
```
maschad, line 12:
```suggestion
language tools while keeping error strings out of on-chain artifacts.
```

IGI-111:
I deliberately didn't want to use "derive", but I'm not opposed to renaming the trait to `Error` and finding a better name for the type annotation.

Maybe `#[error_type]`?

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
+- Feature Name: `abi_errors`
+- Start Date: 2024-11-18
+- RFC PR: [FuelLabs/sway-rfcs#43](https://github.com/FuelLabs/sway-rfcs/pull/43)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
```
maschad, line 4: Should there be a corresponding issue created on the sway repo?

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Error types
+
+Error types are special enums that allow developers to signal that each variant
+is associated with an error message.
+
+An error type implements the `AbiError` trait, but it is not recomended that you
```
maschad, line 48:
```suggestion
An error type implements the `AbiError` trait, but it is not recommended that you
```

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+The recommended way of doing this is to use the the `panic` intrinsic.
+
+`panic` is a compiler intrinsic that will
+produce a unique revert code for each of its invocations and populate the ABI
+specification file with a list of such codes and corresponding information about
+where it was invoked and with what kind of arguments.
+
+If all you want is an error message, you can panic with a string literal:
+
+```sway
```
maschad, line 97: From a consumer perspective if we have to accommodate string literal panics as well as error type panics one could for see inconsistent error handling patterns emerging. My preference would solely Error types.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+It also pollutes the namespace with yet another reserved word.
+
+And ultimately it moves more complexity to the compiler rather than let the user define it.
+This is a shared problem with the error type annotations.
+
+# Rationale and alternatives
+
+[rationale-and-alternatives]: #rationale-and-alternatives
+
+Despite the drawbacks of moving more special behavior the compiler, it has to
```
maschad, line 388:
```suggestion
Despite moving more special behavior to the compiler, Sway must shoulder this complexity since it lacks sufficiently powerful metaprogramming features that would allow developers to define their own language features of this caliber.
```

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+When generating ABI files, the compiler should populate the `"metadataTypes"`
+section accordingly and include a `"errorMessage"` field in the components of
+marked enums.
+
+## `panic` intrinsic
+
+The `panic` intrinsic has two modes which determine what kind of fields are
+produced in the (new) `"errorCodes"` section of the ABI file.
+
+Either it is used with a `str` litteral (or a const-evaluable such value) and
```
alfiedotwtf, line 337: - litteral
+ literal

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+# Future possibilities
+
+[future-possibilities]: #future-possibilities
+
+In the future we may want to include more metadata for panic invocations or
+error types.
+
+We may also allow error strings to be format strings so that you may use your
+enum's values in the error message.
```
alfiedotwtf, line 450: Is there a risk that if we don't do format strings from the beginning then adding them later won't be backwards compatible?

alfiedotwtf:
Looks good and makes sense. Fix the typo and I'm happy to approve :)

kayagokalp (approved):
Looks great to me! This will be a big step forward in error handling.

Other than two spelling typos shown in:

https://github.com/FuelLabs/sway-rfcs/pull/43/files#r1857256227
https://github.com/FuelLabs/sway-rfcs/pull/43/files#r1857482020

It seems good to go to me!

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+Letting the errors be populated directly in the bytecode would be a lot simpler,
+but it is not desirable because it extends the size of the binary for no gain.
+
+Moreover, the question of the location of error strings can be asked. Why
+put them in the ABI specification file? ABI files are soon to benefit from
+a registry that will allow users to easily fetch them, they already contain
+type metadata that can be extended, and they ultimately are our equivalent to
+headers, encompassing all the required information to call a contract and get
+a meaningful anwser. Errors are part of making this answer meaningful and it is
```
JoshuaBatty, line 406:
```suggestion
a meaningful answer. Errors are part of making this answer meaningful and it is
```

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+be able to use error message information to decode revert codes and error types
+that are directly returned.
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+Using a special intrinsic is the main drawback of this solution for a number of reasons.
+
+It muddles how instrinsics are identified since unlike other compiler intrinsics
```
JoshuaBatty, line 373:
```suggestion
It muddles how intrinsics are identified since unlike other compiler intrinsics
```

JoshuaBatty:
Really great work on this.

rfcs/0014-abi-errors.md
ironcev: I agree here. I don't see a reason why not to migrate.

rfcs/0014-abi-errors.md
IGI-111: I can see a purely additive upgrade path from this to format strings: you just record the format strings as normal and change the interpretation from the SDK. We could also add a boolean field right next to them to say they are format strings.

There is a small risk that strings that contain format specifier (i.e.:`{}`) would be affected, but with a new field to mark them this can be done without breakage I think.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+Despite the drawbacks of moving more special behavior the compiler, it has to
+shoulder this complexity in Sway since we do not have sufficiently powerful
+metaprogramming features such that developers may be able to define their own
+language features of this caliber.
+
+Given the benefits of having standardized errors, these tradeoffs seem
+acceptable, so long as we allow the solution to be extended in the future to
+be more user defined. The `AbiError` trait is the main vector of such future
+changes in this case.
```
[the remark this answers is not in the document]
IGI-111: As stated previously, I think I'll just switch this to `Error`. Adding "sway" to any sway type is redundant.

rfcs/0014-abi-errors.md
zees-dev: Agree with this; at the end it may come down to what is most optimal (gas-wise); for example on ethereum explicit errors are generally more gas-efficient than reverts

rfcs/0014-abi-errors.md
ironcev: Hmmm, I find the "minimally burden" better explaining the intention :smile: We want the error strings to be available at the end (somehow, via tooling), but not to burden the bytecode with them.

rfcs/0014-abi-errors.md
ironcev: This comes at the end, when the RFC is accepted.

rfcs/0014-abi-errors.md
ironcev: On one side I agree and like the idea of a uniform approach. On the other side, insisting only on `Error` would create a terrible boilerplate for cases (often cases), when a single error description is just enough.

Instead of just having `panic("This is my issue.");` devs would need to do all the ceremony:
```
#[abi_error]
enum DummyError {
   #[error("This is my issue.")]
   Dummy: (),
}

panic(DummyError::Dummy);
```

Honestly, I think we will soon end up with much more inconsistencies around the choice of `Dummy` word :smile: 

That's why I think supporting `str` is a good tradeof.

rfcs/0014-abi-errors.md
ironcev: Assuming we support it later on, means we will support a syntax similar to `thiserror`: `{}` etc. Theoretically, one might use such characters before as a plain text and cause a breaking change afterwards when they get a special meaning.

One way would be to reserve certain pattern in error strings for future improvements but I see that as too restrictive.

I would rather handle this via breaking changes. We are building a tool right now, `forc migrate` that will support in migrations. One day we introduce format strings we warn devs of a potential breaking change in their existing error strings.

rfcs/0014-abi-errors.md
zees-dev: Thinking through again, the argument of simply panic'ing out is also just as valid; especially for quick iteration

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+put them in the ABI specification file? ABI files are soon to benefit from
+a registry that will allow users to easily fetch them, they already contain
+type metadata that can be extended, and they ultimately are our equivalent to
+headers, encompassing all the required information to call a contract and get
+a meaningful answer. Errors are part of making this answer meaningful and it is
+typical for error messages to live in headers.
+
+If we do not specify some kind of error standard, debugging any sort of complex
+Sway error will remain extremely difficult and supporting Sway contracts may be
+a nigh impossible task.
```
ironcev, line 411: Having a "nigh" in a sentence is nigh impossible not to interpret as a typo at first :smile:  
```suggestion
an almost impossible task.
```

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+It's unclear how much of the existing special error codes produced by the
+standard library (such as `FAILED_ASSERT_SIGNAL`) should be migrated to use this
+mechanism or should remain special behavior.
+
+The extent of the SDK integration of this feature is also quite open ended. But
+implementation of this should give a good enough base to decide how much support
+we want and to extend it in the future.
+
+# Future possibilities
```
ironcev, line 442: One additional nice future possibility to mention explicitly. The proposed approach fits greatly in a possible language expansion for the recoverable errors, namely, having the propagating question mark operator `?` like in Rust.

ironcev (commented):
Great proposal! Straightforward to implement and brings a huge benefit. This is a kind of low hanging fruit I wish we identify more.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Error types
+
+Error types are special enums that allow developers to signal that each variant
+is associated with an error message.
+
+An error type implements the `Error` trait, but it is not recommended that you
```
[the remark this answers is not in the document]
IGI-111: Ultimately the issue is that we need some way of telling if a type is an error and to enforce that any such type:
* is an enum
* has an error message annotation for all of its variants

I guess we could do away with the marker and just have this be enforced by the impl of the marker type? I'm not sure if that's more intuitive or not.

@tritao I'm curious what you think here.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+```sway
+fn do_something(self) -> Result<(), MyError> {
+  // ...
+  return Err(MyError::A); // user will know that this means "error A"
+  // ...
+}
+```
+
+## `panic`
```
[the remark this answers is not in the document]
IGI-111: >I do not understand how the name of the intrinsic, if it is prefixed or not, makes any difference when embedding the error site information?

It doesn't make a difference per se. But it breaks a rule (all intrinsics start with __) and so it behaves surprisingly. I'd like to avoid surprise.

I can see how using a special syntax element like you're suggesting could be less surprising.

As for naming, I'm not really for using revert precisely because it's so widespread and having it as a keyword would pollute the namespace. Panic is less common and it actually has a Rust-like meaning here.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+```sway
+panic(MyError::A);
+// ...
+panic(MyError::B((1024, 42)));
+```
+
+Panicking with an error type will do two things: log the error so that a
+corresponding log receipt is produced, and revert with an error code that
+corresponds to the invocation location.
```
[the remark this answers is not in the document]
IGI-111: There's no real way we can do a backtrace at compile time since you don't know where you're going to be called from.

So yeah it's more about tracking user errors and having some kind of consistent error message than providing full causation information.

I'm open to suggestions on how we could do something like this in a later change, however.

rfcs/0014-abi-errors.md
maschad: Thanks for the clarification @IGI-111 and @ironcev , agree on both points so resolving this.

maschad (commented):
Great work on this @IGI-111 

Not strong opinion on this [suggestion](https://github.com/FuelLabs/sway-rfcs/pull/43#discussion_r1857263982) so feel free to resolve if you don't think it makes the intention clearer.

I'd give my explicit approval but compiler level changes are beyond my scope 😅 

I approve in theory though 😄

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
ironcev: ## TL;DR;
- We need marker traits in general.
- We can unify input to `panic` via marker trait.
- I am not sure if we can have a useful `Error` trait that can actually have implementation.

## The need for marker traits in general

As mentioned, we will need marker traits in general. E.g., in order to express the needed concepts explicitly, without "magic", the Storage 3.0 RFC calls for (work title) [Serializable marker trait](https://github.com/FuelLabs/sway-rfcs/blob/7a1cb3fd429c79e4c318f7675e04fb479d63bf90/files/0013-configurable-and-composable-storage/sway-libs/storage/storage_box.sw#L6).

Also, note that we already assign properties to types, but unlike explicitly by assigning them marker traits, we analyze them on usage site within the compiler and check if they satisfy certain properties. E.g., `AbiEncode` cannot be auto-implemented for types that have pointers. The above link to the proposed (work title) `Serializable` marker trait explains how that trait represent the property, among others, of not having pointers. Having that marker trait would lead to the rule stated explicitly in terms of the type system. E.g., `AbiEncode` is auto-implemented for all types that are marked with `Serializable` marker trait.

This are just some examples why I see the need for marker traits in general.

So yes, assigning them to types is a part of the semantic analysis. When analyzing a type we mark it with all the appropriate marker traits.

In the case of enums that are errors, there is no heuristics in this decision. As per RFC, only those enums that are attributed with `#[error_type]` are marked with the `Error` marker trait.

I see here more issues in naming details, like, shouldn't it be then `core::marker::Error`, etc. But all these details and conventions can be defined when defining the general approach to marker traits in Sway.

The most important feature of Sway marker traits I see, is, that unlike a few of Rust marker traits, they can never be implemented by developers. They are always exclusively assigned by the compiler. (In Rust, for some, it is possible to `unsafe` implement them.)

## Unifying input to `panic` via marker trait `Error`

If we go for this approach, we would define the `Error` marker trait, as: any type that can be accepted by the `panic`. Note that this does not exclude it from being use in `Result`.

An interesting consequence of this is, that `str` can also be marked by the `Error` marker. This then simply means, we can pass `str` to `panic`. There is no special rule for `str` in regards to `panic`. This actually makes `panic` unified. Strictly speaking, the remark that @maschad had about mixing `str` and enums wouldn't be there any more. `panic` accepts `T: Error` and`str` is also marked with the `Error` trait. No exceptions.

## Is a non-marker `Error` actually possible?

Above I've sketched something like:

```
const trait Error {
    fn description(&self) -> str;
}
```

This makes sense at the first look, but in practice, the `Error` instance passed to `panic` must not and will not be a constant. Meaning, in a general/regular case we cannot const evaluate the `description()` as defined above. E.g., to expand on the example from the RFC, all of this will work as per RFC and if the `Error` is just an empty marker trait:

```
pub fn main(a: u8, b: u64) {
  match a {
    4 => panic(MyError::B((b, a))),
    5 => panic(MyError::B((a.as_u64(), b.as_u8()))),
    _ => panic(some_non_const_evaluation_of_my_error(a, b))),
  }
}
```
Note that none of the inputs to `panic`s above is const-evaluable. And it doesn't need to be. For the proposal of the RFC to work, to generate the expected ABI fields, _the compiler only needs to know the type of the provided `Error` instance_. The wiring of the logged instance to the actual error message is done by the tooling later on, as explained in the RFC.

In other words, I don't see how the sketched `const trait Error` implemented by a developer would work in the above example.

The more I think about it, the more I see the approach proposed in the RFC as the right one. The error messages are fixed compile time constants attached to enum variants, not to their particular instances. In the second step, we will allow them to be format strings that can reffere to the logged instance parts, but the calculation of the actual string will be done by the tooling, and not by the compiler.

I hope this all explains why I am for defining marker traits in general and starting here with the marker trait. And even perhaps staying on it long term :wink:

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
ironcev: After giving it a second thought, I am also now for more general `panic` over blockchain specific `revert`. `panic`ing is the semantics of error handling, `revert`ing is an implementation detail in the case of running Sway code on blockchain. It is open for different implementations in case of potentially other runtime environments that we were discussing, while keeping the same error handling semantics of an irrecoverable error.

To the name and nature of `panic`, if it is intrinsic, it should be `__panic`. This conveys the "special handling" and lowers, or even completely removes, the element of "surprise". But it's against the overall nature of intrinsics. They are meant for low-level code, libs etc. We don't want them polluting high level code like contracts. But `__panic` should and will be a regular guest in high-level code. To have `__panic` sticking out in high-level code I definitely don't like, and would, therefore, strike this name through.

Which leads us to the proposed `panic` as either "hidden" intrinsic that looks like a regular function call, or to a new language element, `panic` expression.

Having another one special case (intrinsic without `__`) that also hides the special compiler handling I dislike more then introducing a new expression type. That's the major reason I incline (slightly?) more towards the `panic` expression.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
ironcev: @tritao This is an interesting approach and I see it worth exploring further. My first feeling is that it could work and be very helpful in the debug mode, during the local development. It would give a development experience close to that in "regular" programming. Stacktraces in case of panics etc.

But I see the cost as too high for deployed code.

> There is a very small overhead with this.

I have serious doubts that this would be the case. E.g., right now, the cost of just calculating return addresses during function calls took 2k of 11k of bytecode in a predicate we analyzed. We don't have a notion of stack frames at all on the VM side, function calls are only jumps etc. etc. Manually building that information in memory and keeping track of it at runtime, I cannot imagine that it can be done without exploding the bytecode size. Unless we get specialized VM opcodes. Then of course there is the gas cost attached to it.

But combining the two approaches, this I can definitely imagine:
- The approach defined in this RFC is used for deployed production code. The combination of the compiler generated info and well designed API is indeed lacking the stack trace but still gives rich information for easier analysis of production errors.
- The approach in the comment is used for local development and debugging.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+
+Instead, to create a new error type, use the following syntax and annotations on
+an enum:
+
+```sway
+#[error_type]
+enum MyError {
+  #[error("error A")]
+  A: (),
+  #[error("error B")]
```
sdankel, line 59: Will there be some way to include the error parameters in the error message? Or, will the parameters only be surfaced to the user in the log receipts?

We could support the same syntax as `thiserror` like so:

```sway
  #[error("error B first param: {0}, second param: {1}")]
  B: (u64, u8),
```

sdankel:
Great work, this is going to be a huge improvement! Just left a couple of suggestions/questions.

sdankel (commented):
Great work, this is going to be a huge improvement! Just left a couple of suggestions/questions.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+    "18446744069414584325": {
+        "pos": {
+          "file": "main.sw",
+          "line": 18,
+          "col": 10
+        },
+        "logId": "4933727799282657266",
+        "msg": null,
+    }
+  }
```
sdankel, line 285: These four all have the same `logId` with `msg: null`, the only difference being the position in the file. The first two are `MyError::A` and the second two are `MyError::B(..)`. I can't see how, from this specification, the SDKs would be able to determine which error variant corresponds to these error codes.

Also, this looks like every instance of `MyError::A` will have a different code, which I'm not sure is desirable. For example in unit tests, it would be nice to know that `MyError::A` will always have the same code, so that we can use `should_panic = <code>` to assert that `MyError::A` was thrown.

On that note, it would be even better if we can make `should_panic` support the literal type like `should_panic = MyError::A`, and `should_panic = MyError::B(10, 10)`, but that might be out of scope here. To support it in the future, though, we'd need to lay the groundwork by having a consistent error code per error variant type within a given sway program.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
sdankel: I agree that we'll want a full stack trace for the debugger, and that the stack trace should come from the VM.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
IGI-111: Yeah I agree, the fact it's special does lean towards a special expression, even as that increases complexity slightly.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
IGI-111: I think we all agree that stack traces would be nice but that we're not to burden the resulting final binaries with overhead.

This probably is beyond this particular RFC, I just want to make sure that we don't block ourselves from that by implementing this.

rfcs/0014-abi-errors.md
```diff
@@ -0,0 +1,456 @@
...
+defined types and error strings don't affect the final bytecode size.
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Error types
+
+Error types are special enums that allow developers to signal that each variant
+is associated with an error message.
```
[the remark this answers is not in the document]
IGI-111: All encodable types are supported. Not just tuples, this is just how the Sway syntax works for multivalued enums.

The only limitations are:
* it has to be an enum
* it has to implement `Error`
    * which means it has to implement `AbiEncode`

rfcs/0014-abi-errors.md
IGI-111: There isn't, but this is listed in the possible extensions.

The reason I didn't include this directly here is that we don't actually have format strings in Sway at all yet. And you would want the syntax for this to be the standard one you use everywhere.

rfcs/0014-abi-errors.md
IGI-111: >I can't see how, from this specification, the SDKs would be able to determine which error variant corresponds to these error codes

The SDK doesn't have to figure out which variant from the code because this information is already available in the log: the encoded error contains the enum discriminator.

>every instance of MyError::A will have a different code

This is deliberate. There's no other way to know where the invocation is, and I think this information is important.

The VM can't do this because it does not have access to information about the source code.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
IGI-111: I think this is an interesting discussion but is ultimately off topic since it's an implementation detail.

I agree with most of your points in the utility of having marker traits (even marked as such as a specific concept) but I don't think we should burden the spec with those details.

Suffice to say, a trait should exist so that we can check the type correctness of `panic` invocations, and this is the most expedient solution. Future evolutions of the trait to eliminate some of the magical syntax are, well, future considerations.

rfcs/0014-abi-errors.md
ironcev: > it would be even better if we can make `should_panic` support the literal type like `should_panic = MyError::A`

This would be great usability! And as suggested even having not only the variant name, but also concrete parameters. I think this is all possible within the proposed in this RFC. In the sense that all the building blocks and needed information is there.

Since the code will be assigned by the compiler and visible only in the ABI, I would see the `should_panic = <code>` as a fallback, ideally never needed to be used.

IGI-111: I'm going to close this since the info is available, but I do think using that feature of naming the errors for tests would be a very useful addition.

rfcs/0014-abi-errors.md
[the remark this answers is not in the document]
sdankel: It would be good to include examples of types other than tuples, then.

PR #44, opened 2025-03-10
RFC for const fn.

esdrubal:
[Rendered](https://github.com/FuelLabs/sway-rfcs/blob/esdrubal/const_fn/rfcs/0016-const-fn.md)

### DIFF
--- a/rfcs/0016-const-fn.md
+++ b/rfcs/0016-const-fn.md
@@ -272,27 +272,38 @@ fn sub(x:u64, y:u64) -> u64 {
     x - y
 }
 
-fn main() {
+fn __entry() {
+    ...
+}
+
+fn __const_entry() {
     let A = add(1,2);
     let B = sub(A, 2);
 
-    let A_ptr = __addr_of(&A);
-    let A_size = __size_of_val(A);
-
-    let B_ptr = __addr_of(&B);
-    let B_size = __size_of_val(B);
-
-    __log(raw_slice::from_parts::<u8>(A_ptr, A_size)); // Outputs const A value
-    __log(raw_slice::from_parts::<u8>(B_ptr, B_size)); // Outputs const B value
+    __const_evaluated(0, __addr_of(A));
+    __const_evaluated(1, __addr_of(B));
 }
 ```
 
-All the const evaluated values for the original program are returned using log.
+The whole program will be compiled into IR just once, and will contain both the real
+`__entry` and the `__const_entry`.
 
-The returned log data contains a `Vec<u8>` obtained by using the ABIEncode.
+For the first time `DCE` will run with `__const_entry` as root. Eliminating everything
+that is not needed for the constant evalutation.
 
-Using VM evaluation we can store the generated ABIEncoded byte array into the
-final bytecode of the program.
+This optimized IR will be compiled and run inside the VM. Each constant will be notified
+using the intrinsics `__const_evaluated` which will notify the compiler using an `ecal`.
+
+Compiler will then run a `compacting mark and sweep` algorithm using the calculated consts
+as root. All `marked` memory will then be compressed, and all surviving pointers will be updated.
+
+After that all the survived memory will end up in the `data section` of the binary, and constants
+will be just pointers to this new memory.
+
+This will be "impossible"  to be done with `raw_ptr` and `raw_slice` and types are not 
+defined, and so the compiler will not be able to `mark` all the necessary memory.
+
+### VM evaluation performance
 
 To compile the example above it takes 0.76 seconds including core, without core
 it takes less than 0.02 seconds to compile. To bootstrap the VM and run the
@@ -300,6 +311,15 @@ script it takes less than a millisecond.
 Thus the overhead for doing `const fn` evaluation with the VM should be around
 **tens of milliseconds**, which is negligible compared to compiling a program
 with std which takes around 2.5 seconds.
+
+## Configurables
+
+Today configurables do through ABIEncode/AbiDecode pass. Which not only increase
+the final binary size, but also imposes a gas cost for the mere presence of configurables,
+even when they are not used.
+
+This new approach has the potencial to make `const` and `configurables` initialization really
+zero cost.
 
 ## Enforcing constness rules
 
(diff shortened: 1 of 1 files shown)


xunilrj:
Another "const context" is array size/string array size:

```sway
let a = [0u8; N + 1]; // N + 1 must be a const
let a: str[N + 1] = "abc"; // N + 1 must be a const
```

xunilrj:
> This allows const fns to use heap types internally but restricts types in the data sections to non heap types.

And what about strings? We would not be able to concatenate strings, for example.

```
const NAME: str = "Daniel";
const MSG: str = "Hello" + NAME;
```

rfcs/0016-const-fn.md
```diff
@@ -0,0 +1,406 @@
...
+handling heap type such as `Vec<T>` and function such as `fn push(ref mut self, value: T)`.
+
+A `const fn` with a `ref mut` parameters is callable inside other regular and `const fn`s but
+cannot be used directly in other `const contexts` such as constant items.
+
+## Final binary
+
+When a `const fn`is not called at runtime the final binary outputted by the compiler
+won't have any opcodes related to the computation of `const fn`s.
+
```
jjcnn, line 357: Presumably this is handled by the dead code elimination phase? The UX for this scenario needs to go hand-in-hand with any warnings or whatever when a `const fn` is never called in a const context.

jjcnn (changes requested):
Generally this is perfectly fine when it comes to the semantics of `const fn`, but I wouldn't mind seeing a bit more focus on the UX. What is the user presented with if something goes wrong during compile-time evaulation of a `const fn`? What constitutes unreachable code for `const fn`s? That sort of thing.

ironcev (changes requested):
This review contains my major concerns regarding the proposed VM-based evaluation and writing the resulting content to the data section. The major concern is writing to the data section and skipping the IR representation of the constants.

No matter which eval-execution approach we take (VM, internal interpretation of AST, internal interpretation of IR), IMO, the first three remarks below are a hard requirement for `const fn` and Sway's const-eval in general. 

## `const fn`/eval must be an integral part of the IR optimization pipeline

Writing the VM output to the data section (directly, or compacted or however, but to the data section, skipping the IR) is **exactly what we don't want to have**. **We want to have the eval results as IR constants (`ValueDatum::Constant`), to be able to use them in various IR optimizations**, as it is the case now.

E.g., many evals will result in types like `u64`. **We ideally don't want to have those types in the data section.** We want existing optimizations to put them into immediates or promote them to registers, etc.

Moreover, the RFC proposes that "When a `const fn` is called with the same arguments in different places a unique data section const entry should be used." This is, again, **exactly what we don't want to have**. We want to support const deduplication as we have it now.

E.g., here is a more complex example. The real power of the `const fn` feature is not in having more freedom in expressing declaring `const`s but rather in eliminating runtime expressions in general.

**It must be possible to do const evaluation several times in the IR pipeline, and not only once in the `__const_entry`.**

E.g., in this example:

```sway
fn some_non_const_fn(a: u8) {
       // ... Some code. ...
      let x = some_other_non_const_fn(a, 111);
      // .... More code. ...
}

fn  some_other_non_const_fn(x: u8, y: u8) -> u8 {
       if some_const_fn(22, 33) > 100 {
             // A lot of code.
       } else {
             some_other_const_fn(y)
       }
}
```
Const-evaling `some_const_fn(22, 33)` to something less then `100` will result in eliminating `// A lot of code` branch, leaving only `some_other_const_fn(y)`, which makes `some_other_non_const_fn` eligible for inlining, which exposes it to the caller context where the `y` argument is actually a constant `111`, which again allows us to const eval the inlined `some_other_const_fn(111)` call, effectively completely eliminating the `some_other_non_const_fn(a, 111);` call and replacing it with a constant. Which again can enable more optimizations.

How we structure the IR opt passes to get the best out of const fn eval with acceptable compile time effort is a separate topic. But the major requirement, that const eval results in a `ValueDatum::Constant` that can be used in IR optimizations is, IMO, a hard requirement.

In short, if we just write the VM results to the data section, bypassing the IR and IR optimizations, we will end up with the const-eval mechanism that hinders even the current optimizations, instead of enabling powerful new ones.

## `const fn` must be compatible with storage configuration

Even if we accept the negative impact of writing VM results to the data section on the IR optimizations, as described above, that proposed approach will not work with storage configurations.

**Storage configurations (RHS of the `=` operator in the storage field declaration) are not compiled to bytecode at all, not to data section or anywhere else.** Instead, they are used by the compiler only at compile time to generate slot JSON files. Slot JSON files are generated based on the const-evaled `ValueDatum::Constant`s that are then assigned to storage slots by the compiler.

In other words, **whatever we get as a result of const-evaling the storage configuration expression (RHS), from the VM or whichever evaluation approach we use, this result must be available to the compiler as `ValueDatum::Constant`, to be able to generate slot JSON out of it**.

Note that in the Storage 3.0, both the exact storage slots and the content will be const-evaluated from user defined code (currently it is only the configuration, RHS). Which means both must be available to the compiler and will never end up in the resulting bytecode. They will both be a temporary compile-time artifact.

## Support for slices

Storage 3.0 was one of the major drivers for `const fn` and there we will need support for `const fn` evaluated slices. More specific, the ability to concatenate slices within a const-eval and to return them. The current proposal allows manipulation of heap types during the const-eval execution in VM, but not returning the heap types (types containing pointers and references). While we can have this restriction in general, we will need to support returning slices.

Although, even with the current proposal of VM-based execution, I don't see why continuous parts of memory couldn't be returned (and stored as `ValueDatum::Constant` as argued above).

## Error handling

My conviction was, that with `const fn` we will get a better error handling around const evals. Currently, we just display "Could not evaluate initializer to a const declaration." without any hint what exactly in the chain, and where, went wrong.

I expected to have a backtracking in the const-evaluation that will show the compilation path that lead to the exact point where the error occurred.

If we outsource the evaluation to the VM, we can, after the VM reverts for whatever reason, know which of the `__const_evaluated` calls has failed and which constant cannot be evaluated. But the only error we will be able to show will be the same as now - something went wrong and it's up to you to guess what. Which, in case of intensive use of `const fn`s which we expect, can be very tricky.

Note that, because any of the const evaluation in the `__const_entry` can fail and we need to know which one did, the actual structure of the `__const_entry` will need to be:

```sway
fn __const_entry() {
let A = add(1,2);
__const_evaluated(0, __addr_of(A));

let B = sub(A, 2);
__const_evaluated(1, __addr_of(B));
}
```

Here is one personal experience, an infinite loop caused by not increasing the loop counter. It was difficult to figure it out even when having access to VM output. If it is somewhere in the const-eval chain, nailing it down would be mission impossible.

## Complexity, performance, and tight coupling to VM compared to alternative

These are all obvious concerns.

The complexity of generating the `__const_entry` and the overall mechanics of execution and callbacks. The transformations of the original tree after the `__const_entry` is executed is also not trivial.

IMO we need to elaborate more on performance impact. Can we have the benchmark program attached as a part of this RFC in `files`? Can we have a general PoC that will measure the overall performance impact and not only the one of starting the VM? E.g., we need to compile down the `__const_entry`, get it run, get the data back via callbacks, etc.

Because of all this, **I would love to see the alternative method of execution, inside of the compiler as we have it now, elaborated in more detail**, even if we go with VM-based in the end.

In particular, what are the disadvantages of that approach and how relevant they are, e.g.,:
- no ASM
- encoding VM specific behavior for e.g. overflows
- ???

What are the advantages?
- performance (?)
- state of the art error handling
- simplicity of implementation

What is actually missing for e.g. Storage 3.0 which we see as the most complex case of using const fns?
- a lot of expressions already const-evals, we just need to formalize the `const fn` syntax, 
- support for slices (can be done with dedicated intrinsics, similar to ABI encoding).
- if `Vec` are based on slices they can also participate in `const fn`
- ...

PR #46, opened 2025-06-10
Updated ABI JSON for errors

petertonysmith94:
Update based on the generated ABI from Sway


ironcev (approved):
Thanks for updating @petertonysmith94!

ironcev:
Closing this PR on the following base:
- The feature is meanwhile fully implemented and properly documented. We expect documentation to be used and not RFC.
- Having RFC diverge from the final implementation is not uncommon.

PR #48, opened 2025-09-19
[WIP] Fuzz attributes

kayagokalp:
wip

### DIFF
--- a/rfcs/0016-fuzz-attributes.md
+++ b/rfcs/0016-fuzz-attributes.md
@@ -1,0 +1,192 @@
+- Feature Name: `fuzz_attributes`
+- Start Date: 2025-09-09
+- RFC PR: [FuelLabs/sway-rfcs#0000](https://github.com/FuelLabs/sway-rfcs/pull/001)
+- Sway Issue: [FueLabs/sway#0000](https://github.com/FuelLabs/sway/issues/001)
+
+# Summary
+
+[summary]: #summary
+
+This RFC introduces parsing support for `#[fuzz]` and `#[fuzz_param]` attributes in Sway to enable fuzz testing capabilities. These attributes allow developers to define fuzz test functions and specify parameters for fuzz testing operations.
+
+# Motivation
+
+[motivation]: #motivation
+
+Fuzz testing is a critical software testing technique that helps discover edge cases, security vulnerabilities, and unexpected behaviors by feeding random or semi-random inputs to functions. Currently, Sway lacks native support for fuzz testing attributes, making it difficult for developers to implement comprehensive testing strategies that include fuzzing.
+
+The introduction of `#[fuzz]` and `#[fuzz_param]` attributes will:
+- Enable systematic fuzz testing of Sway functions
+- Provide a standardized way to define fuzz test parameters
+- Improve code quality and security through automated testing of edge cases
+- Align with modern testing practices found in other programming languages
+
+# Guide-level explanation
+
+[guide-level-explanation]: #guide-level-explanation
+
+## Basic Fuzz Testing
+
+The `#[fuzz]` attribute marks a function as a fuzz test. This function will be executed with randomly generated inputs to test for unexpected behaviors:
+
+```sway
+#[fuzz]
+fn fuzz_my_function(input: u64) {
+    // Test logic that will be executed with random u64 values
+    let result = my_function(input);
+    // Add assertions or checks here
+}
+```
+
+## Parameterized Fuzz Testing
+
+The `#[fuzz_param]` attribute allows you to specify parameters for more controlled fuzz testing:
+
+```sway
+#[fuzz]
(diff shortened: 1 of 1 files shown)


rfcs/0016-fuzz-attributes.md
```diff
@@ -0,0 +1,192 @@
...
+
+## Usage Constraints
+
+- A function cannot have both `#[test]` and `#[fuzz]` attributes
+- The `#[fuzz]` attribute expects no arguments
+- Multiple `#[fuzz_param]` attributes are allowed per function
+- Only one `#[fuzz]` attribute is allowed per function
+- `#[fuzz_param]` supports the following parameters:
+  - `name`: The parameter name to configure
+  - `iteration`: Number of test iterations
```
JoshuaBatty, line 64: Just thinking would it make more sense to define this on the `#[fuzz]` attribute? 

so 

```sway
#[fuzz, iteration = 1000]
```

kayagokalp:
# Fuzzing Implementation Discussion Summary

**Related RFC:** https://github.com/FuelLabs/sway-rfcs/pull/48

## Context

Kaya is preparing to make fuzzing functional for Sway and needs to finalize the semantics. An RFC has been initialized to discuss and decide on the implementation approach.

## Participants

- **Kaya** - RFC Author, implementing fuzzing functionality
- **Daniel Leite** - Advocating for library-based approach
- **Igor** - Providing balanced perspective on both approaches

---

## Key Points by Participant

### Kaya's Position

- **Initial Approach:** Started with attribute-based fuzzing (inspired by rstest)
- **Implementation Plan:** Keep executor logic in `forc-test`, limit compiler changes to parser only
- **Future Vision:** Plans include random fuzzing initially, with potential for concolic execution later
- **Open Question:** Is there a downside to attaching this closer to the compiler (in forc-test with compiler probes), other than release cycles?
- **Release Cycle Note:** `forc-test` and other crates in sway repo already share the same release cycle disadvantage
- **Flexibility:** Open to changing direction to library-based approach if consensus favors it

### Daniel Leite's Position

- **Main Concern:** Why use attributes instead of library-based implementation?
- **Proposed Library Syntax:**
  ```rust
  #[test]
  fn fuzz_with_params() {
      for (input, seed) in Fuzz::<(u64, Fuzz::Seed)::new(1000) {
          ...
      }
  }
  ```
  Or with ranges:
  ```rust
  #[test]
  fn fuzz_with_params() {
      for (input, seed) in Fuzz::<(Fuzz::U64InRange<0, 100>, Fuzz::Seed)::new(1000) {
          ...
      }
  }
  ```
- **Key Arguments:**
  - Library approach possible because macros exist (though Sway doesn't support them like Rust)
  - Compiler should only handle what libraries cannot (e.g., RNG seed generation)
  - Library-based approach avoids waiting for compiler releases for new features
  - Reduces need for compiler changes
- **Acknowledgment:** Agrees that advanced features like concolic execution may need more compiler help, but suggests finding "bare minimum" compiler requirements

### Igor's Position

- **Core Issue:** Sway doesn't support macros like Rust does, which is why attributes require compiler changes
- **Developer Experience Concern:** Library-only approach without attributes would create suboptimal DX with annoying boilerplate
- **Balanced View:**
  - Shares Daniel's concerns about compiler coupling
  - Worried that library-only approach will frustrate developers
  - Acknowledges flexible library could offer more and be less constrained
- **Compiler Support Assessment:** Only attributes definition and checking needed in compiler; execution logic stays in `forc-test`
- **Current Position:** Indecisive between lib and attribute approaches
- **Hybrid Approach Consideration:** "Lib + attribute where compiler lowers to lib usage" seen as more effort than plain attributes with execution in `forc-test`
- **Scope Expansion Suggestion:**
  - Should brainstorm PBT (Property-Based Testing) and Concolic testing as part of this RFC
  - At minimum, PBT should be included as it's close to fuzzing in look-and-feel
- **RFC Title Suggestion:** Change from attribute-specific to general "Fuzzy Testing in Sway"
- **Community Engagement:** Should move discussion to GitHub and actively reach out to community, especially the Telegram team that implemented attribute-based fuzzing

---

## Open Questions

1. **Attribute vs Library Approach:** Which provides better balance of DX vs maintainability?
2. **Compiler Coupling:** What is acceptable level of compiler involvement?
3. **Release Cycle Impact:** Is shared release cycle with `forc-test` acceptable?
4. **Scope of RFC:** Should this RFC cover only fuzzing, or also PBT and concolic execution?
5. **Community Input:** What does the Telegram team (who implemented attribute-based fuzzing) think about library approach?

## Next Steps

1. Move discussion to GitHub (on RFC PR)
2. Reach out to community for broader input
3. Consider RFC scope expansion to include PBT
4. Decide on implementation approach (attribute, library, or hybrid)
5. Update RFC title if scope expands beyond attributes


Summarized by claude from a private chat so that we can move the discussion to public

PR #51, opened 2026-04-16
SemanticDefinitions

xunilrj:


xunilrj:
This seems something that the `SemanticDefinition` can also help: https://github.com/FuelLabs/sway/issues/7605

If we think about it... there is no much different checking something on types after they become concrete, with checking something on const generics if they become concrete.

I see a similar api here:

```rust
sd.when_const_expression_is_non_changeable(array.length, |_, _, final_value| {
    if index >= final_value {
        handler.emit_err(...);
    }
});
```

